PR #4, opened 2023-03-24
Add the Pattern-Based Collections Proposal

gitamohr:
### Description of Proposal
Here we propose to add extensible pattern-based collection support to OpenUSD, for use in `UsdCollectionAPI` and other domains, drawing inspiration from CEL and Solaris.

Please provide any feedback on this pull request.  You can view the [rendered proposal in my fork](https://github.com/gitamohr/USD-proposals/blob/main/proposals/pattern-based-collections/README.md).

### Contributing


flord:
This will fix the biggest missing feature of USD in my opinion. Being used to Solaris, I really like the syntax. Thanks!
I was wondering about the composition order. Are the pattern based collections evaluated before composition or after? If it's before, if means they can only contain prims defined in the current root layer and all sublayers. If it's after composition, they could contain prims that are defined in higher layers, leading to very bad surprises when someone uses "//*" to assign a material in the modeling layer and someone else loads the entire layerstack. 
I also assume these collections don't propagate outside of reference arcs, so "//*" used in an asset will not propagate to the entire shot.

gitamohr:
Thanks for the feedback, @flord .  As it stands the intent is to handle path patterns analogously to how we do relationship targets and attribute connections, so your suspicion is correct: `//*`, used in an asset will be scoped to that asset, not propagate to the entire shot.

nvmkuruc:
I've looked through the proposal and had some questions.

## Scoping
The document refers to the predicate domain as being determined by `UsdCollectionAPI`. However, if it's a new attribute type, can any schema add expressions assuming the same available predicates? Are there any examples of what the scene description would look like?
## Relationship to relationships
It's interesting that expressions are being proposed as an additional attribute type instead of as an extension to relationships or a new property type. It feels like many of the uses cases for expressions are currently in the domain of relationships. I'd also expect relationships to be the natural target for users looking to archive / freeze / flatten an expression.
## Stage Flattening
Flattening usually preserves the composed result, even if the arcs are not preserved. Does this suggest that for robustness expressions must necessarily be evaluated and flattened as well?
## Impact on Notice Listeners
Tooling dependent on the computed result of expressions may have to do a lot more evaluations of these expressions. Are there API / Utilities that can be provided to help tools minimize / optimize expression matching for a notice?
## Scene Graph Instancing
What's the expected behavior with respect to scene graph instancing, prototypes, and proxies? It seems necessary that proxies and not prototypes are considered in the predicates, as you don't want to deal with behavior changing as instancing switches flip on and off (or a prototype root path changes). However, do compact expressions complicate efficient scene description, rendering, and traversal? An instance's efficiency can be procedurally undone with expressions. Are there thoughts on expressing collections and expressions in ways that align with instancing?
## Abstract Prims
Are there use cases for abstract prims being expression targets that warrant inclusion in scene description? While setting a material binding directly on a class would affect its inherited prims, I don't believe there's the expectation that targeting an abstract prim as a target of material binding affects its inheriting prims (or any API that helps achieve this). For C++/python traversal predicates, being able to include / exclude abstract makes sense (if for no other reason than user interface display / filtering of abstract prims), but what's the reasoning for considering abstractness in scene description expressions?
## Purpose / Attribute
Is there a case for why purpose is privileged as the only attribute in predicates? In general, I'd be hesitant about opening the door to any attributes being allowed in predicates of expressions at the Usd level, at least initially.
## Application State Predicates
Are there cases where application state should affect / filter predicates? For example, when evaluating material bindings, if I'm not rendering absent prims, I might want to avoid computing the expression on an absent prims. However, if I enable visualization of absent prims, I'll presumably need / want to compute those bindings. These predicates don't feel like they can be embedded in scene description yet seem potentially necessary to minimize. These predicates may be different than / complimentary to the ones expressed in scene description.
## Schema Predicates
One feature I wish relationships had was being able to annotate at the schema level validation rules. That is to say-- I wish could say `rel<Mesh>` or `rel<Property>` to denote that this relationship is intended to target meshes or properties. Has there been any consideration to adding restrictive expression predicates at the schema level? As a designer of a schema, I could imagine being thoughtful about the performance implications of expressions or wanting some additional validation assistance. For example, a material binding expression predicate could be setup to only target prims in an imageable hierarchy, short circuiting traversal of material networks when looking for members. This would reduce the burden on users to properly setup such an expression.
## Model Hierarchy Predicates
It's been my experience that the `IsModel` / `IsGroup` language has been confusing. It's been my experience that when people reach for `IsModel()`, they mean `IsModel() && !IsGroup()` or `IsKind("component")`. I've had the most success explaining model hierarchy in terms of assembly, components, and subcomponents and encouraging users to think of `IsModel()` as something like a `IsAnElementInAHierarchyOfModels()`. While I understand that this would be symmetrical with the first class USD traversal predicates, I suspect that privileging the `group` / `model` kinds in predicates may lead to incorrect usage when most users probably are looking for the `assembly` / `component` kinds. The predicates `kind:model` and `kind:group` would still allow intentional filtering by these.
## Predicates and Validation
One interesting aspect of the proposal is that predicate rules can be used to optimize matching. However, this is dependent on the scene description being "correct". This could lead to some interesting and surprising transformations. If I have a `component` asset and I reference it into another `component` asset without making it a `subcomponent`, that's not "valid model hierarchy" today, but the consequences are quite minimal. But let's say that referenced `component`'s material bindings are determined via an expression that considers `component`-ness. Once it's referenced into another component model, the referenced component is no longer a valid component. Repairing the model hierarchy and converting to `subcomponent` upon reference would also break the expression (as it's no longer a `component`). I suspect that successful user experience's with expressions will be tied to validation and correctness tooling as well as strong guidelines and documentation to nudge users from developing conflicting asset and expression structures.
## Recursive Match Syntax
The `//` recursive match seems like an expensive operation, but that complexity feels somewhat masked. It takes up less characters / screen real estate than the cheaper `/*/` non-recursive match. In the spirit of form following function, I'd advocate that some characters exist between the `/`s to indicate that expense. `/**/` would be symmetrical with the globstar syntax. `/.../` would be another pattern to consider.
## Specifier
The `specifier` predicate was unclear to me. Is that for matching against `uniform` vs. `varying` attributes?
## ListOps
ListOps have proven themselves to be an effective way of expressing accumulated elements. A formulation of expressions that may `append`, `prepend` or `remove` elements from weaker opinions seems like a natural fit. Was that considered instead of the `%` character?
## Evaluation Context Mapping
Are there any thoughts as to whether or not expressions evaluation could / should be deferred into hydra for performance reasons? If some hydra component prunes large portions of the scene graph, you could imagine warning to elide expression evaluation.

spiffmon:
Just wanted to respond quickly to two things:

- **Model Hierarchy Predicates**:  There's a years-languishing ticket somewhere to add `IsComponent` and `IsAssembly` queries efficiently (i.e. tracking the state with cached prim flags) to UsdPrim so that they could be added to traversal predicates, and so that we can try to stomp out the `IsModel() && !IsGroup()` pattern.  Agree 100% the current state of model hierarchy queries is confusing.  Would you want to take a crack at that? :-) 
- **Evaluation Context Mapping**: indeed, and that has been the plan from the start.  Given that the evaluation rules will be well-defined, we expect Hydra 2 to be evaluating material-bindings and light-linking collections in the Hydra space, not on the UsdStage, using the same engine that USD would, but fed from different data.

gitamohr:
> ### Scoping
> The document refers to the predicate domain as being determined by `UsdCollectionAPI`. However, if it's a new attribute type, can any schema add expressions assuming the same available predicates? Are there any examples of what the scene description would look like?

Yes, and the scene description is just a string containing the expression.

> ### Relationship to relationships
> It's interesting that expressions are being proposed as an additional attribute type instead of as an extension to relationships or a new property type. It feels like many of the uses cases for expressions are currently in the domain of relationships. I'd also expect relationships to be the natural target for users looking to archive / freeze / flatten an expression.

A new kind of property that is neither an Attribute nor a Relationship would be a massive upheaval to the entire software stack and far too costly.

It would be possible to extend relationships to do this, but that would mean at least an extra cost to pay when looking at any relationship ever to see if it had "pattern-based" targets vs list-op targets. We would also have to decide on semantics when both the pattern-based specification and list-ops are authored. Also, this only works in the pure USD-object domain; we would still need another way to encode pattern-based collections of objects from other domains with other predicate functions relevant to those domains. Introducing a new attribute type gives us a single way to handle both USD-objects and other domains, and we only pay for it when it's used.

> ### Stage Flattening
> Flattening usually preserves the composed result, even if the arcs are not preserved. Does this suggest that for robustness expressions must necessarily be evaluated and flattened as well?

No, we will just write the composed pattern expressions to the flattened result.

> ### Impact on Notice Listeners
> Tooling dependent on the computed result of expressions may have to do a lot more evaluations of these expressions. Are there API / Utilities that can be provided to help tools minimize / optimize expression matching for a notice?

I'm not sure what you mean, could you give an example?

> ### Scene Graph Instancing
> What's the expected behavior with respect to scene graph instancing, prototypes, and proxies? It seems necessary that proxies and not prototypes are considered in the predicates, as you don't want to deal with behavior changing as instancing switches flip on and off (or a prototype root path changes). However, do compact expressions complicate efficient scene description, rendering, and traversal? An instance's efficiency can be procedurally undone with expressions. Are there thoughts on expressing collections and expressions in ways that align with instancing?

It is in principle possible to make the evaluation instancing-aware -- my initial thoughts are essentially a memoization approach where we record matches under instance prototypes.  That said, the first implementation may not include this optimization.

> ### Abstract Prims
> Are there use cases for abstract prims being expression targets that warrant inclusion in scene description? While setting a material binding directly on a class would affect its inherited prims, I don't believe there's the expectation that targeting an abstract prim as a target of material binding affects its inheriting prims (or any API that helps achieve this). For C++/python traversal predicates, being able to include / exclude abstract makes sense (if for no other reason than user interface display / filtering of abstract prims), but what's the reasoning for considering abstractness in scene description expressions?

I don't know of any offhand, but to me that's not a good reason to leave a possibly useful predicate out.  I don't see any harm in allowing predicates that test for abstractness, do you?

> ### Purpose / Attribute
> Is there a case for why purpose is privileged as the only attribute in predicates? In general, I'd be hesitant about opening the door to any attributes being allowed in predicates of expressions at the Usd level, at least initially.

Ah yes good call -- I'll discuss removing this.

> ### Application State Predicates
> Are there cases where application state should affect / filter predicates? For example, when evaluating material bindings, if I'm not rendering absent prims, I might want to avoid computing the expression on an absent prims. However, if I enable visualization of absent prims, I'll presumably need / want to compute those bindings. These predicates don't feel like they can be embedded in scene description yet seem potentially necessary to minimize. These predicates may be different than / complimentary to the ones expressed in scene description.

Yes that makes sense.  Internally we anticipate our main initial uses of these expressions to actually be in the "hydra space" and make use of hydra-specific predicates that are not discussed here.  You are free to write and use your own predicates; as mentioned you can supply whatever you like to the evaluation engine.  And you can choose to encode them in scene description, or not.  We do anticipate encoding hydra-specific predicates in scene description, but not on the collection api schema.

> ### Schema Predicates
> One feature I wish relationships had was being able to annotate at the schema level validation rules. That is to say-- I wish could say `rel<Mesh>` or `rel<Property>` to denote that this relationship is intended to target meshes or properties. Has there been any consideration to adding restrictive expression predicates at the schema level? As a designer of a schema, I could imagine being thoughtful about the performance implications of expressions or wanting some additional validation assistance. For example, a material binding expression predicate could be setup to only target prims in an imageable hierarchy, short circuiting traversal of material networks when looking for members. This would reduce the burden on users to properly setup such an expression.

Generally this falls under the heading of "Validation", which Presto/Mf has, and to my knowledge we will be embarking on a project to add this to USD soon. And yes, as mentioned in the proposal, the code that builds the evaluator gets to specify the domain of paths by supplying what amount to traversal functions.  So you would be free to prune material networks in your evaluator.

> ### Model Hierarchy Predicates
> It's been my experience that the `IsModel` / `IsGroup` language has been confusing. It's been my experience that when people reach for `IsModel()`, they mean `IsModel() && !IsGroup()` or `IsKind("component")`. I've had the most success explaining model hierarchy in terms of assembly, components, and subcomponents and encouraging users to think of `IsModel()` as something like a `IsAnElementInAHierarchyOfModels()`. While I understand that this would be symmetrical with the first class USD traversal predicates, I suspect that privileging the `group` / `model` kinds in predicates may lead to incorrect usage when most users probably are looking for the `assembly` / `component` kinds. The predicates `kind:model` and `kind:group` would still allow intentional filtering by these.

I agree with you and Spiff on this; specifically we'd want to add the query to the core, and then we could add the predicate to the pattern language.

> ### Predicates and Validation
> One interesting aspect of the proposal is that predicate rules can be used to optimize matching. However, this is dependent on the scene description being "correct". This could lead to some interesting and surprising transformations. If I have a `component` asset and I reference it into another `component` asset without making it a `subcomponent`, that's not "valid model hierarchy" today, but the consequences are quite minimal. But let's say that referenced `component`'s material bindings are determined via an expression that considers `component`-ness. Once it's referenced into another component model, the referenced component is no longer a valid component. Repairing the model hierarchy and converting to `subcomponent` upon reference would also break the expression (as it's no longer a `component`). I suspect that successful user experience's with expressions will be tied to validation and correctness tooling as well as strong guidelines and documentation to nudge users from developing conflicting asset and expression structures.

I generally agree with this but I'm not sure how this affects this proposal.

> ### Recursive Match Syntax
> The `//` recursive match seems like an expensive operation, but that complexity feels somewhat masked. It takes up less characters / screen real estate than the cheaper `/*/` non-recursive match. In the spirit of form following function, I'd advocate that some characters exist between the `/`s to indicate that expense. `/**/` would be symmetrical with the globstar syntax. `/.../` would be another pattern to consider.

We've bikeshedded this a bunch. Folks coming from Katana/CEL have strong feelings about this particular syntax.

> ### Specifier
> The `specifier` predicate was unclear to me. Is that for matching against `uniform` vs. `varying` attributes?

No it's about testing prim specifier: "def", "over", "class".

> ### ListOps
> ListOps have proven themselves to be an effective way of expressing accumulated elements. A formulation of expressions that may `append`, `prepend` or `remove` elements from weaker opinions seems like a natural fit. Was that considered instead of the `%` character?

I think you mean the `%_` specification of the "weaker pattern"?  To my mind this provides a simpler and more powerful way to augment or modify a set than list ops can do.  Since these are sets, "prepend" and "append" don't really make sense, but you can add and remove things simply.  For example, `/new/item %_` is like "prepend", and `%_ /new/item` is like "append".  Doing `%_ - /removed/item` is like "remove".  Moreover you get to use all of the set algebra operations. But please let me know if you are thinking of a list-op style operation that would not be possible with this proposal.

> ### Evaluation Context Mapping
> Are there any thoughts as to whether or not expressions evaluation could / should be deferred into hydra for performance reasons? If some hydra component prunes large portions of the scene graph, you could imagine warning to elide expression evaluation.

Yes definitely.  The USD core won't do any pattern testing or evaluation "automatically".  For patterns that operate on the hydra space they will certainly be evaluated by hydra and operate only on its own domain (with its own predicates).

nvmkuruc:
> Yes, and the scene description is just a string containing the expression.

That's helpful to hear. Some of my questions were predicated on that a new attribute type meant a new `SdfValyeTypeName` and scene description would look something like this.

```
expression my_expression = </my/pattern>
```

gitamohr:
Oh it will be a new value type, but the right hand side will be the string expression, so like:
```
pathExpression myExpr = "//nuts //bolts"
```

tymonpitts:
### Where I'm Coming From
I'm in the midst of adding pseudo "pattern-based collections" at my studio for the purpose of material binding. Users will be able to provide a pattern which we'll then bake out to static collections under the hood. We're not using native Solaris nodes for this because we need a DCC-agnostic solution. This proposal was really good timing since I was starting to think about what syntax to use right when this came out!

###Matching Properties - Yes Please!
We definitely need the ability to test uniform attribute values so adding my vote for that feature since the proposal mentions it might not be in the first implementation. We don't need support for time-varying attributes, just uniform.

###Matching Properties - Syntax Ambiguity
Looking at the "Matching Prims by Testing Properties, and Matching Properties" section of the proposal, it's ambiguous what the actual syntax would be for different operators and various types of attributes. For example, The proposal only mentions this example for the syntax: `.radius{value:default:closeTo:0}`

- I'm assuming `value:default` is checking against the result of `attr.Get(Usd.TimeCode.Default())`?
- How close is `closeTo`? 6 decimals?
- How do you check exact equality instead of approximate equality? `equals` instead of `closeTo` (e.g. `.intAttr{value:default:equals:0}`?
- How do you check the value of a string type? I'm assuming `.stringAttr{value:default:equals:"myValue"}` or would there be no quotes? The same ambiguity exists for the `predicate(arg1, keyword=value, ...)` syntax. Would `value` in this case be `keyword="myValue"` or just `keyword=myValue`?
- Can you do other string comparison operations like `startswith` and `endswith` or would you create a new predicate function for that? (I don't necessarily need this, just wondering)

###Matching Properties - Syntax Verbosity
I'd also say the syntax is very verbose for property matching:
- Why not something simpler like `.intAttr=1`? The `closeTo` operator could just be `~=` (e.g. `.radius~=0`). 
- Is there something other than value matching that I'm not thinking of which would necessitate the more verbose `{value:...}` syntax? If not then why not just `{default:closeTo:0}`?
- As for "default", could we just say that expressions don't support matching time-varying data and so can only match the default value? That way we can simply omit `default` from the syntax (e.g. `{value:closeTo:0}` instead of `{value:default:closeTo:0}`. 

###Expression Parsing API
For my use case I'd like to present a nice UI for the user to assemble their expression. Something with better UX than a simple text field. This UI would need to be able to populate itself with an existing expression so it would be really nice to have a first class API to parse that expression. Without this I'll need to define my own PEG and use my own PEG library to parse it (which is what I'm doing now). While that's entirely possible for me to do, it'll be cumbersome to match USD's internal PEG 1 to 1 and stay up to date with any changes to USD's internal PEG.

spiffmon:
On syntax verbosity around attribute values, there are at least two things we're thinking about

1. Some generators of scene description seem to generate a "uniform" value by authoring everything at frame/time zero (or some other single timeCode).  There may be value in not penalizing such scene description to be unmatchable, so we've talked about defining "the value" (or some other predicate function) as "evaluate at `UsdTimeCode::EarliestTime()` "
2. There could also be a case for evaluating at a specified timeCode ?

tymonpitts:
Then perhaps something like the following would be more generally simple and only verbose when you need the verbosity?

- `.intAttr=0` uses `Usd.TimeCode.Default()`
- `.intAttr=0{TimeCode:EarliestTime}` uses `Usd.TimeCode.EarliestTime()`
- `.intAttr=0{TimeCode:2.5}` uses `Usd.TimeCode(2.5)`

It just seems cumbersome to complicate the 90% usecase to satisfy the 10%

nvmkuruc:
Thanks Alex for the response. I'm putting together some additional thoughts on the model hierarchy questions which might help explain why I think they're relevant to the proposal, but they are more general and don't need to continue in this thread.
--
Regarding relationships, I'm still not sure I have a complete picture of how relationships and path expressions relate. The place where I'm coming from is that I suspect that pipelines will want to be able to cache path expressions for archival or performance purposes.  Relationships seem like the obvious target for cached expressions. To me it sounds like there will be code at some point that will be balancing / swapping / combining relationship representations and expression representations. Is the idea that this logic will just be built into collections?  And that the expectation is that most schema builders are choosing between collections and relationships (and between path expressions and relationships)?
-- 
Regarding notices, rereading the proposal I /think/ the answer to my question is that probably some combination of `SdfPathExpressionEvaluator` + `SdfPathTable`.
--
Regarding `/**/` vs. `//`. If you're open to me amending my suggestion, maybe it's helpful to support both `/**/` and `//`. This might be easier for users not familiar with CEL to pick up. It also might be easier to document noting `//` as an alias for `/**/` instead of `//` as a replacement `/**/`. I understand there may be other costs to having two representations, so certainly disregard if this doesn't resonate.

meshula (approved):
ready to land/review/iterate

phuchau1989 (approved):

PR #6, opened 2023-04-05
Propose `PickAPI` as a specification for picking customization

nvmkuruc:
### Description of Proposal

Asset creators may want to customize the picking behavior of their assets for performance or other practical reasons.  This proposal outlines a specification for pick visibility and retargeting rules for imageable prims.

[Proposal](https://github.com/NVIDIA-Omniverse/USD-proposals/tree/PickAPI/proposals/PickAPI)

### Contributing

### DIFF
--- a/proposals/PickAPI/README.md
+++ b/proposals/PickAPI/README.md
@@ -1,0 +1,184 @@
+# PickAPI
+## Overview
+Picking refers to the selection of one or more prims through an interactive rendering of the scene. Picking encompasses both point-and-click and sweep selection interactions. Picking is often accelerated through the use of the rendering toolkit, such as a pick buffer or a bounding volume hierarchy.
+
+This document proposes a specification for customized prim picking behavior to benefit the interchange of a variety of scenes and assets in USD.
+
+## Motivation
+### Interactive Scenes
+A designer of an interactive scene might want to limit picking inside of a room to a set of specific scopes. While objects may not be pickable, the designer may still choose to have them occlude picking of other objects (ie. Walls shouldn't be pickable, but picking shouldn't select objects behind walls).
+
+### Performance
+Asset builders may want to improve selection speed for consumers by removing expensive to render prims from pick buffers. Downstream consumers understand and expect that these objects will both not be pickable and not occlude picking (ie. Rendered fur on a character may be invisible to picking, allowing the underlying surface to be picked).
+
+### Retargeting
+Asset builders may be aware of important scopes made up of multiple gprims, say a group of prims representing a door. They understand that downstream users selecting the door knob are generally trying to rotate the entire door and would benefit from directly selecting the rotatable root prim.
+
+## Proposal
+To describe high level, customized primitive pick behavior, provide a `PickAPI` schema in `UsdUI`. This schema describes two key functions [visibility](#visibility) and [retargeting](#retargeting).
+
+This proposal distinguishes between customization and [application pick modes](#relationship-to-application-pick-modes). Pick modes are user interface options that apply to the entire scene and cannot be practically encoded on every prim in the scene.
+
+### Visibility
+```
+uniform token pick:visibility = "inherited" (allowedTokens = ["inherited", "invisible"])
+```
+`pick:visibility` controls whether or not the object should be considered visible with respect to any picking operation. Objects 'invisible' to picking should be excluded from any picking calculations (and acceleration structures like BVHs or pick buffers). Prims should only be included in the pick buffer if they are imaged, taking into account `visibility`, `purpose`, and any other viewport render settings.
+
+Because picking is often accelerated through the rendering toolkit, it's important to describe picking on its terms.
+        
+If the `PickAPI` is applied to multiple sites in a hierarchy, `pick:visibility` has pruning semantics similar to the `visibility` attribute. If a prim is invisible to picking, so are its descendants.
+
+To make an object _unselectable_ but participate in picking operations as an occluder, users must use [retargeting](#retargeting-for-pick-occluders).
+
+`pick:visibility` may not be animated.
+
+In this current proposal, sending geometry to a pick buffer that is not actively imaged is currently out of scope. One hypothetical use case would be interactive picking of volumetrics. A user might provide a mesh to be used as the pick target for a dense volumetric cloud. This proposal doesn't provide a path for including non-imageable geometry in picking because the semantics become somewhat confusing and complicates implementations leveraging rendering pipelines.
+
+### Retargeting
+
+```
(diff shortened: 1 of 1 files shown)


proposals/PickAPI/README.md
```diff
@@ -0,0 +1,192 @@
...
+
+If the `PickAPI` is applied to multiple sites in a hierarchy, the retargeting behavior will be defined by the nearest applied ancestor. This is notably different than the pruning semantics of `pick:visibility`. As specified, retargeting could be set to `none` in a descendant, overriding any ancestral retargeting opinions.
+
+Targets will usually include one or more prims, but may include one or more properties. If a relationship is targeted, standard relationship forwarding should be applied. If an attribute is targeted, that attribute should be selected if meaningful, otherwise the attribute's parent prim. There is no way to use the `PickAPI` to retarget selection to a relationship.
+
+#### Retargeting for Pick Occluders
+If targets is empty, no prims will be selected, and the hierarchy will act as a matte occluder. Utilities (say `UsdUIPickAPI(prim).OccludeOnly()`) can be used to explicitly set `pick:visibility` to `inherited`, `pick:retargeting` to `replace`, and `pick:targets` to an empty list.
```
spiffmon, line 57: Instead of "replace to empty" to get matte occlusion, might it be clearer to have `pick:retargeting = "matteOcclusion"` or something like that?

proposals/PickAPI/README.md
```diff
@@ -0,0 +1,192 @@
...
+
+### Retargeting
+
+```
+uniform token pick:retargeting = "none" (allowedTokens = ["none", "replace"])
+rel pick:targets
+```
+These attributes control whether or not to forward picking to alternative targets. The default behavior (`none`) is to not forward picking. However, if `replace` is selected, all descendant prim selections will be replaced with `pick:targets`.
+
+If the `PickAPI` is applied to multiple sites in a hierarchy, the retargeting behavior will be defined by the nearest applied ancestor. This is notably different than the pruning semantics of `pick:visibility`. As specified, retargeting could be set to `none` in a descendant, overriding any ancestral retargeting opinions.
```
spiffmon, line 52: Given that retargeting is non-pruning, why not make the `pick:visibility` non-pruning, as well, to have the same semantics for all picking aspects?  While `imageable.visibility` _is_ pruning, `imageable.purpose` is not.  If we called it `pickable` instead of `visibility`, I don't think there would be any confusion.  I believe this would also provide more flexible in setting up mattes.

proposals/PickAPI/README.md
```diff
@@ -0,0 +1,192 @@
...
+            uniform token pick:visibility = "inherited"
+            uniform token pick:retargeting = "replace"
+            rel pick:targets = <../Glass>
+        }
+    }
+}
+```
+
+## Questions
+* This proposal does not address support for `UsdGeomSubset`. Visibility and purpose are currently not supported by subsets, so it may be hard to support the visibility attribute. However, one can imagine a desire for using face sets to describe pick retargeting.
```
spiffmon, line 186: Retargeting based on face-sets is a clear need for character models, for sure.  Subsets are the obvious answer that would fit elegantly with this proposal... modulo the issue you called out here.  Making Subset Imageable has been discussed and pretty much agreed to on usd-interest (the not-fully-resolved issue is that we have no way of rectifying a different `purpose` for a Subset from its parent Mesh, in current "face" usage).  More recently, though, we have been planning on migrating visibility and purpose out of Imageable and into VisibilityAPI, which would remove the "technical" impediment completely.

proposals/PickAPI/README.md
```diff
@@ -0,0 +1,192 @@
...
+* Should pick visibility and and pick retargeting be two separate APIs? One could imagine tools providing facility for controlling pick visibility but not supporting retargeting. One could also imagine prims like `UsdGeomSubset` not supporting pick visibility but supporting pick retageting. One challenge with this formulation is identifying whether the visibility or retargeting schemas are responsible for describing pick occluders without creating ambiguity.
+* Is `UsdUI` the right place for viewport behavior schemas?
```
spiffmon, line 188: It feels similar to `UsdUINodeGraphNodeAPI` to me, in that it's embedding extra information to guide a certain class of user interactions that have no bearing on "core computation or rendering"?

spiffmon (commented):
A few initial thoughts... still chewing on some of the other questions at the bottom - nice writeup, @nvmkuruc !

proposals/PickAPI/README.md
nvmkuruc: I considered that.  A version of this proposal had something like

```
uniform pick:retargeting = "none" (allowedTokens = ["none", "matte", "replace"])
```

It occurred to me that `replace` with empty targets would necessarily be equivalent to `matte` so I dropped the explicit option in the proposal in favor of their being one and only way to specify something. I similarly dropped a `self` mode in favor of just targeting `<.>`.  I can definitely see the reasoning for the explicit formulation, perhaps just for the `matte` case?

proposals/PickAPI/README.md
nvmkuruc: This is a good question.  I considered at one point having a few modes for `pick:visibility`. One hypothetical use case was to provide a pick proxy for a hard to pick prim like a volume.  I thought this made things more confusing, as it required a mode that would override / ignore other visibility state as you wanted it to only render in the pick buffer (but only sometimes ignore the other visibility state).

One could argue that this "pick buffer only" rendering could or should be an extension to `purpose`.

I ultimately settled on strict pruning visibility as I suspected the simplest formulation would be the most successful path to introduce new behavior into rendering contexts.

proposals/PickAPI/README.md
spiffmon: That's a good concern.  Maybe it's just me, though, but I'd consider making `replace` with empty targets be an error condition, because one _could_ reasonably (I think?) assume not `matte` behavior but instead just "unpickable with passthrough" behavior?

nvmkuruc:
I've updated the proposal to incorporate some of Spiff's questions and added a speculative section on `GeomSubset` handling.

I'm considering making some modifications to the proposal to better flesh out the story with respect to instancing. Would be interested if anyone has any thoughts on the complexities instancing introduces into a picking specification.

meshula (approved):

meshula:
Matt, I've marked this "approved" so we can merge it, with Draft status I think. Could you do the commit signing fandango please?

With respect to instancing, isn't it true that in general, you'd want a redirect to the instancing prim itself, and in rare cases, presumably via management at the code level, you might want to redirect towards the prototype itself? I'm curious what your thoughts on instance handling were.

meshula:
@dsyu-pixar Could I ask you to add an index entry for PickAPI in Draft status please?

PR #8, opened 2023-04-24
Add expression syntax to Stage Variable Expressions proposal

sunyab:
### Description of Proposal
This is an update to the Stage Variable Expressions proposal primarily adding details about the intended expression syntax and (initial) functions to be provided.

The rendered view of this document can be seen [here](https://github.com/PixarAnimationStudios/USD-proposals/blob/b6c56931f255a7e5c3ae3d2fd53a4ba06e3a5876/proposals/stage_variable_expressions/README.md).

### Contributing

### DIFF
--- a/proposals/stage_variable_expressions/README.md
+++ b/proposals/stage_variable_expressions/README.md
@@ -120,10 +120,10 @@ primSpec.variantSelections['modelVariant'] = '`if(eq(${SHOT}, "01"), "shot_01_va
 ## Stage Variables
 
 Stage variables are the only scene description that expressions are allowed to
-refer to. These variables may be strings, bools, or ints and are stored in a
-dictionary-valued layer metadata field named ``stageVariables``. Stage variables
-may themselves be expressions, so that common logic can be factored into a
-single location for convenience and brevity.
+refer to. These variables may be strings, bools, integers, or lists of these
+types, and are stored in a dictionary-valued layer metadata field named
+``stageVariables``. Stage variables may themselves be expressions, so that
+common logic can be factored into a single location for convenience and brevity.
 
 As the name implies, stage variables are stage metadata, meaning they must be
 authored on either the root or session layer of a stage or in the root layer of
--- a/proposals/stage_variable_expressions/README.md
+++ b/proposals/stage_variable_expressions/README.md
@@ -5,6 +5,9 @@ Copyright &copy; 2022, Pixar Animation Studios,  version 1.0
 ## Contents
   - [Introduction](#introduction)
   - [Expressions](#expressions)
+    - [Syntax](#syntax)
+    - [Functions](#functions)
+    - [Grammar](#grammar)
   - [Stage Variables](#stage-variables)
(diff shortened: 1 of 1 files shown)


proposals/stage_variable_expressions/README.md
```diff
@@ -106,24 +126,146 @@ error messages. If we didn't explicitly identify expressions somehow, the only
...
+<double-quoted-string-contents> = [<double-quoted-escaped-char> | <char> | <stage-variable>]*
+<double-quoted-escape-char> = "\" ['"' | "$" | "\"]
+
+; A function is an identifier followed by a comma-separated list of
+; expressions, surrounded by "(" and ")".
+<function> = <identifier> "(" [<expression> ["," <expression>]*] ")"
+
+; A list is a comma-separated list of non-list expressions, surrounded by
+; "[" and "]".
+<list> = "[" [<expression_no_list> ["," <expression_no_list>]*] "]"
```
spiffmon, line 259: In the enclosing "##Expressions" section, we state that we support "One-dimensional, homogeneous arrays of the above types."

But this `<list>` element looks like a heterogeneous container, as defined, and there is no mention of arrays.  Can you clarify, please?

proposals/stage_variable_expressions/README.md
```diff
@@ -94,10 +97,27 @@ def "CrowdCharB" (
...
+# Variant Selections:
+primSpec.variantSelections['modelVariant'] = '`if(eq(${SHOT}, "01"), "shot_01_variant", "shot_other_variant")`'
+```
+
+Expressions are functional and strongly-typed. The supported value types are:
+
+- Integer (limited to the same minimum and maximum values as an int64_t)
+- Boolean
+- Strings (single or double quoted, using "\" as an escape character)
+- One-dimensional, homogenous arrays of the above types.
```
spiffmon, line 118: (sp) homogeneous

proposals/stage_variable_expressions/README.md
```diff
@@ -106,24 +126,146 @@ error messages. If we didn't explicitly identify expressions somehow, the only
...
+; During evaluation, these references will be substituted with the value
+; of the named stage variable. 
+;
+; Strings use backslashes (\) to escape delimiters, including the leading
+; "$" in a stage variable reference.
+<string> = <single-quoted-string>
+         | <double-quoted-string>
+
+<single-quoted-string> = "'" <single-quoted-string-contents> "'"
+<single-quoted-string-contents> = [<single-quoted-escaped-char> | <char> | <stage-variable>]*
```
meshula, line 246: At the enclosing scope, we'll be parsing for back-ticks to mark the expression to be parsed. I think the enclosing parse needs to be smart enough to skip back-ticks in quoted strings, but does need to be smart enough to interpret escaped characters (to know where the end of a string containing an escaped quote ends). Do we say so somewhere? I think my larger question is if alternatively, char is too broad here, and if there should be a narrower definition that only allows escaped characters and unescaped characters necessary to write a legal expression?

meshula (approved):

spiffmon (commented):
This looks great, @sunyab - just a few questions and edits.

PR #9, opened 2023-04-26
Adding "TOC" list of proposals to proposals/Readme.md

dsyu-pixar:
### Description of Proposal

Not a proposal per se, just adding a list of proposals to proposals/Readme.md for easier navigating of current proposals.

### Supporting Materials

N/A

### Contributing

### DIFF
--- a/proposals/Readme.md
+++ b/proposals/Readme.md
@@ -8,7 +8,7 @@ within the proposal.
 
 ## Current Proposals
 
-- [Pattern-based Collections](https://github.com/PixarAnimationStudios/USD-proposals/pull/4) ![Status:Draft](https://img.shields.io/badge/status-draft-blue)
-- [Pick API](https://github.com/PixarAnimationStudios/USD-proposals/pull/6) ![Status:Draft](https://img.shields.io/badge/status-draft-blue)
-- [Stage Variable Expressions](https://github.com/PixarAnimationStudios/USD-proposals/pull/8) ![Status:Draft](https://img.shields.io/badge/status-draft-blue)
-- [Unicode Identifiers in USD](https://github.com/PixarAnimationStudios/USD-proposals/pull/5) ![Status:Draft](https://img.shields.io/badge/status-draft-blue)
+- [Pattern-based Collections](https://github.com/PixarAnimationStudios/USD-proposals/pull/4) ![Status:Draft](https://img.shields.io/badge/draft-blue) ([current PRs and issues](https://github.com/PixarAnimationStudios/USD-proposals/pulls?q=is%3Apr+label%3Ausd-pattern-based-collections+))
+- [Pick API](https://github.com/PixarAnimationStudios/USD-proposals/pull/6) ![Status:Draft](https://img.shields.io/badge/draft-blue) ([current PRs and issues](https://github.com/PixarAnimationStudios/USD-proposals/pulls?q=is%3Apr+label%3Ausd-pickapi+))
+- [Stage Variable Expressions](https://github.com/PixarAnimationStudios/USD-proposals/tree/main/proposals/stage_variable_expressions) ![Status:Draft](https://img.shields.io/badge/draft-blue) ([current PRs and issues](https://github.com/PixarAnimationStudios/USD-proposals/pulls?q=is%3Apr+label%3Ausd-stage-variables))
+- [Unicode Identifiers in USD](https://github.com/PixarAnimationStudios/USD-proposals/tree/main/proposals/tf_utf8_identifiers) ![Status:Draft](https://img.shields.io/badge/draft-blue) ([current PRs and issues](https://github.com/PixarAnimationStudios/USD-proposals/pulls?q=is%3Apr++label%3Ausd-utf8-identifiers+))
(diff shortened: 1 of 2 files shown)


proposals/Readme.md
```diff
@@ -5,3 +5,10 @@ Create new proposals here, in their own directories.
 A Readme.md file should exist in each directory, and may contain 
 the entire proposal, or it may serve as an index to the rest of the materials
 within the proposal.
+
+## Current Proposals
```
dsyu-pixar, line 9: I arbitrarily opted to have list in alpha-order. 

The status badges can be tweaked, or we can ditch them altogether.

meshula (approved):
lgtm!

proposals/Readme.md
```diff
@@ -5,3 +5,10 @@ Create new proposals here, in their own directories.
...
+
+- [Pattern-based Collections](https://github.com/PixarAnimationStudios/USD-proposals/pull/4) ![Status:Draft](https://img.shields.io/badge/status-draft-blue)
+- [Pick API](https://github.com/PixarAnimationStudios/USD-proposals/pull/6) ![Status:Draft](https://img.shields.io/badge/status-draft-blue)
+- [Stage Variable Expressions](https://github.com/PixarAnimationStudios/USD-proposals/pull/8) ![Status:Draft](https://img.shields.io/badge/status-draft-blue)
+- [Unicode Identifiers in USD](https://github.com/PixarAnimationStudios/USD-proposals/pull/5) ![Status:Draft](https://img.shields.io/badge/status-draft-blue)
```
spiffmon, line 14: This proposal (and Stage Variables) have actually been largely committed.  @meshula, @dsyu-pixar , what should our practice be in these "intermediate" states?  Do we **only** link to the published document if/when there are no open PR's remaining?  That would be a bit of a pain because after the first PR is merged, if _another_ PR gets filed (as in the case of the UTF-8 proposal), we need to keep changing the target back and forth from document to PR.

There's also the issue of tracking/collecting all of the PRs and Issues for a "Final" proposal, to access the discussions.  One idea we tossed around was creating a unique label for each proposal and applying it to all Issues and PR's for that proposal, but maybe there are better ways of organizing them.

meshula: Now that you mention it, one of the ideas we talked about originally was that the individual proposals could be organized within the proposals dir as

```
+-- nifty/
        README.md
        proposal.md
        +--- supplemental/
            nifty.usda
            nifty.jpg
```
where the README.md in the proposal would provide a quick intro, a status report, a link to the proposal and supplementals, and then the `nifty` tag would search issues and PRs.

The Toc would then link to the README.md in the proposals dir, and the README could link to a search on tags.

Here's an example for usd-utf8-identifiers, a label I added just now:

https://github.com/PixarAnimationStudios/USD-proposals/pulls?q=is%3Apr++label%3Ausd-utf8-identifiers+

meshula: The idea with the label format is to name it as `usd-myproposal` for usd, `hyd-myproposal` for hydra, and so on.

dsyu-pixar: I like the idea of having a proposal-level README along with labels -- I think this addresses Spiff's concern of having to regularly update the top-level TOC whenever new PRs are created, landed, etc. 

If we still want to show status information in the top-level TOC, I think we'd still have to manually update that, unless there's some clever way in Markdown of pulling that status info from a child README.md (which I don't think there is such a way). I think manually updating this is ok, however, since I feel like the status isn't going to change at the same frequency as new PRs?

spiffmon: I also like the idea of the README + proposal md’s, and love the label-search!  I am not in love with mashing “target component” into the proposal-label, though.  I don’t think it scales, and is restrictive - what if a proposal spans USD schemas, openExec computations, and imaging?  I’d leave the proposal-labels pure, and add `hydra`, (for the future) `openExec`, and (if necessary) `usd` labels.

Agreed with @dsyu-pixar that updatng the badges on the ToC shouldn’t be too frequent or onerous?

meshula: The thing I was wrestling with was whether we need a proposals prefix or not. I figured we needed something to "key" unique proposals. I thought about `p-`, `prop-`, `proposal-`, and `usd-`. Something to distinguish a descriptor `utf8-attribute`, which apply to many things, versus `proposal-utf8-attribute` which is the high level thing.

Maybe we could use the PR# of the initial PR? So the Stage Variables proposal could be labeled `pr7-stage-variables` as the master key? or `rfc7-stage-variables`? `up7-stage-variables` (~~with reference to the Pixar movie~~ standing for USD Proposal)?

meshula: I am kind of fond of `up#-` as a prefix :)

dsyu-pixar: Added an update to the TOC to incorporate the label links. 

The TOC lines are getting kinda long, but they're still readable if you're just looking at the file in a shell/text editor. I made a table-based version as an experiment, and it looks better in the rendered markdown, but worse if you're just looking at the text.

meshula: They look fine to me.

proposals/Readme.md
meshula: I like the badges

proposals/Readme.md
meshula: I think we can start with this. We can revise the labels to have no prefix or a different prefix as a follow up.

phuchau1989 (approved):

PR #10, opened 2023-05-24
Propose self assembling model hierarchy

nvmkuruc:
### Description of Proposal
This [proposal](https://github.com/NVIDIA-Omniverse/USD-proposals/blob/selfAssemblingModelHierarchies/proposals/selfAssemblingModelHierarchy/README.md) advocates for simplifying model hierarchy construction and maintenance through automatic propagation of model `group`-ness.

### Supporting Materials
N/A

### Contributing

### DIFF
--- a/proposals/selfAssemblingModelHierarchy/README.md
+++ b/proposals/selfAssemblingModelHierarchy/README.md
@@ -1,0 +1,208 @@
+# Self Assembling Model Hierarchy
+Copyright © 2023, NVIDIA Corporation, version 1.0
+
+## Goal
+Simplify correct model hierarchy construction through removal of the need for explicit `kind=group` tagging.
+
+## Background
+USD presents "model hierarchy" as a mechanism for efficient traversal of a stage's "important" prims. This importance generally corresponds to the stage's referenced assets, but is pipeline independent, persistent under flattening of arcs, and allows for "casual" use of composition operators without implying "importance" (ie. using internal references to reuse scene description).
+
+Model hierarchy is expressed through a prim's `kind` metadata. It's worth noting that `kind` is an extensible concept, but that won't be addressed in this document as to participate in model hierarchy, they must derive from one of the official `kind`s. `kind` is distinct from and orthogonal to a prim's schema type.
+
+* `component`: This `kind` represents the leaf in model hierarchy traversals. What defines a `component` can be context dependent. In the context of a city, each building may be a `component`. In the context of a room, pieces of furniture may be `component`s. As they represent leaves in the model hierarchy, they _cannot_ be nested.
+* `assembly`: Assemblies aggregate `component` models. In a city of buildings, the city would be an `assembly`. Assemblies may be nested. For example, a city `assembly` may contain multiple district `assembly` prims.
+* `subcomponent`: These prims do not participate in model hierarchy traversals, but indicate important descendant prims of `component` models. `subcomponent` prims may be arbitrarily nested. In a city of `component` buildings, the doors on each building may be "important" to picking and other user interface browsers. `component` models that need to reference other `component` models often will override the `kind` to `subcomponent` to avoid violating model hierarchy rules.
+
+The fourth first class kind is that of `group` which must be explicitly specified on prims that are not `assembly` prims but may contain `assembly`  or `component` prims. `group` acts to bridge model hierarchies on non-"important" scopes that may contain important descendants. For completeness, it's important to note that `assembly` prims are necessarily `group`s because they may contain models (`component` or other `assembly` prims).
+
+Let's look at an example of model hierarchy correctly used.
+```
+def Scope "TriStateArea" (kind = "group") {
+    def Xform "NewYorkCity" (kind = "assembly") {
+        def Scope "Water" (kind = "group") {}
+        def Scope "Bridges" (kind = "group") {
+            def Xform "BrooklynBridge" (kind = "component") {}
+        }
+        def Scope "Tunnels" (kind = "group") {
+            def Xform "LincolnTunnel" (kind = "component") {}
+        }
+        def Scope "Boroughs" (kind = "group") {
+            def Scope "Manhattan" (kind = "assembly") {
+                def Scope "FifthAvenue" (kind = "group") {
+                    def Xform "DepartmentStore" (kind = "component") {
+                        def Scope "GroundFloor" {
+                            def Scope "Housewears" (kind = "subcomponent") {
+                                def Xform "Aisle_1" {
+                                    def Mesh "TopShelf" {
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+```
+> **Note** "GroundFloor" and "Aisle_1" do not have `kind` authored. It would be incorrect to set `kind=group` or `kind=assembly`
+here. As a descendant of `component`, it may only be unset, or a `subcomponent`.
+
+> **Note** There's no relationship between the type of a prim being a `Scope` or an `Xform` and its authored `kind`.
+
+We've described `assembly` prims as being "important" prims generally corresponding to referenced assets. While not provided or mandated by USD, it is likely that pipeline and tooling is supporting versioning and setup of `assembly` and `component` models.
+
+For prims with `kind=group`, importance in model hierarchy is implicit from the fact that they _may_ contain important prims. `group` tagging exists solely so that traversal knows to keep going.
+
+## Problem
+There aren't currently consequences to violating model hierarchy so it happens a lot. Model hierarchy doesn't affect rendering so tools and users may not be properly maintaining it. Even pipelines that attempt to honor model hierarchy may choose to repair model hierarchy only at specific validation points.
+
+What makes this worth addressing now is the new pattern based collection [proposal](https://github.com/PixarAnimationStudios/USD-proposals/pull/4) aims to leverage model hierarchy in its predicates. Model hierarchy will now affect (and potentially accelerate) collection membership computation. **Tagging a prim as an `assembly` or a `component` could change the results of material bindings or light linking.**
+
+Proper maintenance of model hierarchy involves performing hygene up and down the hierarchy. For users to rely on model hierarchy in their predicates, model hierarchy maintenance can't be deferred to a validation script or ignored.
+
+Let's view an invalid version of model hierarchy on the above example.
+```
+def Scope "TriStateArea" (kind = "group") {
+    def Xform "NewYorkCity" (kind = "assembly") {
+        def Scope "Water" (kind = "group") {}
+        def Scope "Bridges" {
+            def Xform "BrooklynBridge" (kind = "component") {}
+        }
+        def Scope "Tunnels" {
+            def Xform "LincolnTunnel" (kind = "component") {}
+        }
+        def Scope "Boroughs" (kind = "group") {
+            def Scope "Manhattan" (kind = "assembly") {
+                def Scope "FifthAvenue" (kind = "group") {
+                    def Xform "DepartmentStore" (kind = "component") {
+                        def Scope "GroundFloor" {
+                            def Scope "Housewears" (kind = "subcomponent") {
+                                def Xform "Aisle_1" {
+                                    def Mesh "TopShelf" {
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+```
+Some scopes (`Bridges`, `Tunnels`) are missing their group tagging. A model hierarchy centric traversal of the stage will miss the `LincolnTunnel` and `BrooklynBridge`.
+
+The automatic fix is obvious-- Find all `component` models and tag any untagged parents with `kind=group`. It seems like USD could just figure this out on its own eliding the need for `kind=group`. However, model hierarchy is cached during composition and not multi-pass. Making prim flag computation multi-pass is too complex a solution to this problem.
+
+## Proposal
+The USD Glossary has this aspirational take on model hierarchy maintenance.
+
+> When assets are published/packaged with model kinds already established, the model hierarchy becomes mostly “self assembling”.
+
+As the above examples demonstrate, a self assembling model hierarchy unfortunately has been complicated by the need for the "glue" `group` kind.
+
+There is prior art for composition "fixing" errant model hierarchy. Consider just the department store. It's not uncommon to see users (incorrectly) apply the `group` kind to all `Xform` and `Scope` prims in a hierarchy.
+```
+def Xform "DepartmentStore" (kind = "component") {
+    def Scope "GroundFloor" (kind = "group") {
+        def Scope "Housewears" (kind = "subcomponent") {
+        }
+    }
+}
+```
+
+The authored `group` kind is ignored with respect to model hierarchy because its parent isn't a `group` or `assembly`. With minimal few trade-offs, this proposal argues that just as incorrect usage can be discarded, _correct `group` usage can be propagated_.
+
+More simply-- **untagged children of groups and assemblies are automatically groups**.
+
+When applied recursively, this simple change would correct most incorrect model hierarchy usage and vastly simplify maintenance. Model hierarchy would be "self assembling".
+
+Let's revise the example, removing sites where explicit authoring of `kind=group` would no longer be necessary.
+```
+def Scope "TriStateArea" (kind = "group") {
+    def Xform "NewYorkCity" (kind = "assembly") {
+        def Scope "Water" {}
+        def Scope "Bridges" {
+            def Xform "BrooklynBridge" (kind = "component") {}
+        }
+        def Scope "Tunnels" {
+            def Xform "LincolnTunnel" (kind = "component") {}
+        }
+        def Scope "Boroughs" {
+            def Scope "Manhattan" (kind = "assembly") {
+                def Scope "FifthAvenue" {
+                    def Xform "DepartmentStore" (kind = "component") {
+                        def Scope "GroundFloor" {
+                            def Scope "Housewears" (kind = "subcomponent") {
+                                def Xform "Aisle_1" {
+                                    def Mesh "TopShelf" {
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+```
+In this example, because `NewYorkCity` is an `assembly` (and therefore a `group`), its descendant `Water`, `Tunnels`, `Bridges`, and `Boroughs` scopes are automatically `group` prims. Propagation would end at the `BrooklynBridge`, `LincolnTunnel`, and `DepartmentStore` `component` models.
+
+(For completeness, it's worth noting that the pseudo-root is automatically a `group`. Propagation will need to be suppressed in that case.)
+
+This proposal has minimal trade-offs, not no trade-offs. Consider this example.
+
+```
+# The "East River Conundrum"
+def Xform "NewYorkCity" (kind = "assembly") {
+    def Scope "Water" (kind = "group") {
+        def Mesh "EastRiver" {}
+    }
+}
+```
+
+The "EastRiver" is geometry just inlined as a mesh into the "NewYorkCity" assembly. It is not violating model hierarchy rules but has a non-`component` `Mesh` as a descendant of a `group`. The trade-offs come from how this case should be handled.
+
+#### Variant #1: `component` prims are still the only leaves
+We can say that it's okay that the "EastRiver" is now considered a `group`. All we've said is that it "might" contain models. The vast majority of prims in the scene graph will still be skipped during model hierarchy searches.
+
+The biggest drawback is when you considered material networks introduced at the assembly level, there could be a large number of prims to traverse. It may not make sense to nudge these materials into `component` models to prune traversal.
+
+#### Variant #2: Introduce a new `auto` kind which enables propagation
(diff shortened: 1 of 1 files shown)


proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+The USD development team proposes that prim flags are considered in concert with composition arcs when constructing prototypes to deal with this ambiguity.
+
+Until that fix is available, the best that can be done in both the current state of the world and with self propagating hierarchies is to encourage users to be intentional about model hierarchy when authoring `instanceable` to avoid violating the hierarchy's continuity.
+
+### Additional API
+
+To further simplify usage of model hierarchy for developers, this proposal also advocates introducing `IsAssembly`, `IsComponent`, and `IsSubcomponent` methods on `UsdModelAPI`, along with associated cached prim flags and traversal predicates.
+
+This proposal also recommends `IsModel` should be deprecated in favor of `IsModelHierarchy` and should be equivalent to (`IsGroup() || IsComponent()`).
```
spiffmon, line 194: I think that would be an improvement, and prevent the UB we get today from `kind = “model”` . Unfortunately, it’ll be tricky for Pixar to make this change given the historical uses of kind in our pipeline… I think that implementing the [“mix-in kinds” proposed on USD-interest](https://groups.google.com/g/usd-interest/c/S8pOpqr52DI/m/yFNNghUBBwAJ) would help us make the transition away from model and derived-from-model kinds in our assets.

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+The USD core API could also provide ranges to iterate explicitly over `assembly`, `component` or `subcomponent` prims, skipping over the intermediate prims. This might be easier to provide in C++20 (with the forthcoming ranges specification) and could be deferred until that point.
+
+## Summary
+This proposal aims to fufill the promise of self assembling model hierarchy as indicated in the USD documentation. If rendering behavior is going to depend on model hierarchy being correctly specified through path expressions, it's important to simplify correct assembly and maintenance of model hierarchy for users and tools.
+
+As the "East River Conundrum" demonstrates, there isn't a trade-off free solution to this problem, but importantly, the trade-offs can be largely mitigated for currently correctly specified model hierarchies and API usage.
+
+## Questions
+* This proposal rules out multi-pass population of prim data. Was that a correct assumption?
```
spiffmon, line 206: We did special work for payloads to avoid two-pass composition, and yes we’d like to avoid two-pass population for similar performance reasons.  Also, the two-pass approach, while avoiding the **traversal** pathology of the “East River Conundrum”, still incurs the “read kind on potentially all prims during population” cost.

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,208 @@
...
+* This proposal hand waives prevelance of invalid model hierarchies as "a lot". Are the cases of invalid model hierarchy ultimately anecodatal and tooling for assembling it strong enough that it doesn't warrant any changes?
```
spiffmon, line 207: We would love to hear feedback on this, also.  By far the most common error in our pipeline is in the “one model references in another model” case, neglecting to override he embedded model’s kind to “subcomponent”, which is not as impactful for the concerns expressed in this document.

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,208 @@
...
+* Should incorrect model hierarchy (`component`, `assembly`, and `group` models under `component` models) be communicated as warnings rather than just suppressed?
```
spiffmon, line 208: Doing so would sacrifice our optimization, and we are hoping that having readily available validation that catches such errors will be sufficient?

spiffmon (commented):
This is a thoughtful proposal with several good ideas.  It seems like the key “auto-propagation-of-group-ness” feature, though, trades a potential correctness problem (which *should* be fairly noticeable in renders?) for much more subtle potential performance problems, and our instincts say the former is preferable to have to deal with?

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+#### Variant #3: Use authored state to control propagation
+Only propagate when `kind` is unauthored. This is effectively the same as `auto` in that "EastRiver" can be tagged with `kind=""` to suppress propagation.
+
+In user interfaces, it may be hard to disambiguate between unauthored (and propagating) and authored (propagating) `kind` state. There would also not be way to author a value to re-enable propagation, but this may be fine.
+#### Recommendation
+This proposal advocates #2 or #3 as being the best choice as they minimize impact to current usage. The only thing that would have to change is `kind=""` would have to be explicitly authored to explicitly stop propagation. The biggest risk to either of these usages is that code sites may be doing explicit `GetMetadata("kind")` calls and comparisons instead of using the `UsdModelAPI`. For clarity, some of these variations may benefit from the concept to author a more explicit model hierarchy block-- say `kind="none"`, but no proposal strictly requires introducing this.
```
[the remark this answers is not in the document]
nvmkuruc: If model hierarchy isn't used at all, I don't /think/ things should change performance wise.  The proposal still requires kind to be explicitly set on root prims.

proposals/selfAssemblingModelHierarchy/README.md
nvmkuruc: Good point on another drawback of multi-pass.

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,208 @@
...
+### Instancing
+Propagation when `instanceable=True` is complicated, as a prim may have multiple parents with different model hierarchy validity. There are actually [bugs](https://github.com/PixarAnimationStudios/USD/issues/2406) with this today.
+
+The USD development team proposes that prim flags are considered in concert with composition arcs when constructing prototypes to deal with this ambiguity.
+
+Until that fix is available, the best that can be done in both the current state of the world and with self propagating hierarchies is to encourage users to be intentional about model hierarchy when authoring `instanceable` to avoid violating the hierarchy's continuity.
+
+### Additional API
+
+To further simplify usage of model hierarchy for developers, this proposal also advocates introducing `IsAssembly`, `IsComponent`, and `IsSubcomponent` methods on `UsdModelAPI`, along with associated cached prim flags and traversal predicates.
```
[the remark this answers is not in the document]
nvmkuruc: Do assembly queries necessarily have to be uncached?  I understand why subcomponent queries are more complicated.

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+For prims with `kind=group`, importance in model hierarchy is implicit from the fact that they _may_ contain important prims. `group` tagging exists solely so that traversal knows to keep going.
+
+## Problem
+There aren't currently consequences to violating model hierarchy so it happens a lot. Model hierarchy doesn't affect rendering so tools and users may not be properly maintaining it. Even pipelines that attempt to honor model hierarchy may choose to repair model hierarchy only at specific validation points.
+
+What makes this worth addressing now is the new pattern based collection [proposal](https://github.com/PixarAnimationStudios/USD-proposals/pull/4) aims to leverage model hierarchy in its predicates. Model hierarchy will now affect (and potentially accelerate) collection membership computation. **Tagging a prim as an `assembly` or a `component` could change the results of material bindings or light linking.**
+
+Proper maintenance of model hierarchy involves performing hygene up and down the hierarchy. For users to rely on model hierarchy in their predicates, model hierarchy maintenance can't be deferred to a validation script or ignored.
```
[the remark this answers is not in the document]
meshula: Would the resolution of this question be to replace the indicated section with a note about preferring correctness problems to speed problems, and that the new validation framework could or must validate MH?

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,208 @@
...
+```
+def Xform "DepartmentStore" (kind = "component") {
+    def Scope "GroundFloor" (kind = "group") {
+        def Scope "Housewears" (kind = "subcomponent") {
+        }
+    }
+}
+```
+
+The authored `group` kind is ignored with respect to model hierarchy because its parent isn't a `group` or `assembly`. With minimal few trade-offs, this proposal argues that just as incorrect usage can be discarded, _correct `group` usage can be propagated_.
```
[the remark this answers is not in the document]
meshula: @nvmkuruc Could you amend with a note about validation?

proposals/selfAssemblingModelHierarchy/README.md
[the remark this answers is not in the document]
meshula: Recommendation has been removed, resolving conversation.

proposals/selfAssemblingModelHierarchy/README.md
meshula: Would the resolution here be to remove deprecation recommendation?

proposals/selfAssemblingModelHierarchy/README.md
meshula: Let's move this question and discussion to an issue in order that the conversation may be marked resolved, and we can move towards landing this proposal in Draft form.

meshula (changes requested):
Most questions are resolved. Can we update the proposal to address the outstanding comments, and therefore progress to landing this proposal with Draft status?

proposals/selfAssemblingModelHierarchy/README.md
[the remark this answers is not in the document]
nvmkuruc: I'll update the proposal that better considers the implications of validation, but I'm not sure I'm drawing the same conclusion. If validation can be used to detect and repair correctness problems, it seems like it could be used to detect and report any performance problems related to how a model hierarchy is structured.

proposals/selfAssemblingModelHierarchy/README.md
nvmkuruc: Renaming `IsModel` to `IsModelHierarchy` doesn't necessarily mean that `kind=model` has to go away. I can update with that note.

proposals/selfAssemblingModelHierarchy/README.md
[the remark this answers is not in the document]
nvmkuruc: I'll update with a broader note about validation.

nvmkuruc:
> Most questions are resolved. Can we update the proposal to address the outstanding comments, and therefore progress to landing this proposal with Draft status?

I pushed a new draft which tries to address to comments.  Thanks @meshula.

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,342 @@
...
+
+Prior to path expressions, delaying validation until `Save`-ing of a stage was perhaps viable. Not necessarily ideal that model hierarchy would be a broken state, but the impact would be limited to a session's interactive features and not imaging. Delaying validation as a post-process could lead to different material binding and light collections. A user saving a stage they are happy with and then validation informing them of errors that change the way the stage renders risks erroding user trust of validation. It also could lead to users avoiding building expressions around model hierarchy features. As model hierarchy is intended to be efficiently cached, relying on adhoc tags or naming conventions could reduce overall performance of path expressions.
+
+Self assembling model hierarchy does not remove the need for validation of model hierarchy. However, it focuses it on the scopes that are changing and reduces the amount cases that can be considered invalid. We enumerate the potential invalid cases below.
```
meshula, line 321:
```suggestion
Self assembling model hierarchy does not remove the need for validation of model hierarchy. However, it focuses it on the scopes that are changing and reduces the number of cases that can be considered invalid. We enumerate the potential invalid cases below.
```

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,342 @@
...
+### Performance of incomplete model hierarchies
+While self assembling model hierarchy does not introduce any new correctness issues, there is the potential for an unterminated hierarchy to do unnecessary reads of the `kind` field. This performance impact in the context of a stage is likely to be measurable but very small (1-2%). Proficient users and maintainers of model hierarchy should never encounter measurable performance issues.
+
+Validation could be used to "optimize" an inefficient instead of "correct" an invalid hierarchy. Under self assembling model hierarchy, it would become acceptable to introduce optimizing terminal kinds when validating in a post-process. This proposal argues that users are likely to favor validators that optimize content without changing imaging behavior.
```
meshula, line 330:
```suggestion
Validation could be considered to "optimize" an inefficient hierarchy rather than "correcting" an invalid hierarchy. Under self assembling model hierarchy, it would become acceptable to introduce optimizing terminal kinds when validating in a post-process. This proposal argues that users are likely to favor validators that optimize content without changing imaging behavior.
```

meshula (changes requested):
@nvmkuruc I have a bunch of mostly typo/grammar changes. If you like them, could you whack "accept suggestion" and if we can close them all out that way, we can go ahead and merge.

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,342 @@
...
+
+### Invalid Case #1: Missing root prim `kind`
+As a concession to performance and "pay for what you use", a hierarchy must opt into model hierarchy at the root prim. Validating and repairing a missing root prim kind has the same risks issues and without automatic group propagation. We suggest that users, tools, and pipelines are willing to treat root prims as special and more willing to carefully maintain and validate their correctness than intermediate scopes.
+### Invalid Case #2: Nested `component` models
+When `component` (or `assembly`) models are nested, the general recommendation is to turn the nested model into a `subcomponent`. Importantly, following this recommendation will not impact path expression matching. Model hierarchy population should have already discarded nested models in both. Converting them to `subcomponent` shores up existing model hierarchy correctness, but does not change the "model"-ness of any ancestors or descendants. (_NOTE-- It's possible that path expressions may match against subcomponents, and so a more conservative validation recommendation may be to convert their kind to the empty string._)
```
spiffmon, line 326: There is no such recommendation for nested `assembly` models - assemblies of assemblies are expected and not uncommon in our pipeline.

Regarding the recommendation about fixing component-under-component to the empty string... despite the fact that it could change pattern evaluation, `component` carries some amount of importance with it, and demoting it to "nothing" seems undesirable... possibly this is one of those validation fixer cases where a choice of fix is needed?

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,342 @@
...
+`auto` becomes `kind`'s new fallback value. The "EastRiver" and other members of the incomplete hierarchy could be explicitly tagged with `kind=""` to preserve current behavior. Propagation for `group` or `assembly` kinds is `group`.
+
+#### Variant #2: Use authored state to control propagation
+Only propagate when `kind` is unauthored. This is effectively the same as `auto` in that "EastRiver" can be tagged with `kind=""` to suppress propagation.
+
+In user interfaces, it may be hard to disambiguate between unauthored (and propagating) and authored (not propagating) `kind` state. There would also not be way to author a value to re-enable propagation, but this may be fine.
+
+#### Variant #3: Use `subcomponent` (or some other kind) to prune traversal
+Despite its name, there are no rules that `subcomponent` prims are descendants of `component` prims. One could use that or some other kind to meaningfully stop traversal. `kind=none` (or `kind=terminal`) has been another suggestion. Even if variant #1 or #2 were adopted, this would still be a valid way to prune traversal, it just wouldn't be the only (or preferred) way.
```
spiffmon, line 287: Given that we are only ever proposing to propagate when the relevant root prim is explicitly tagged as a model container, why not just say that `component` is the preferred terminator?  What's the point in declaring a model-containing root prim if you are not going to add any models into it?  And if it's a case of "some assets/prims have proper component tagging, but others don't, it seems the pattern is pretty clear and someone (or some software) just forgot to properly tag some asset, then let's address the traversal-pruning problem by adding components.  As you argue below, even adding `subcomponent` tagging to prune traversals _could_ change the contents of some collection path expressions, so if we're going to "break" things, let's break them in the most organized way ?

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,342 @@
...
+```
+def Xform "DepartmentStore" (kind = "component") {
+    def Scope "GroundFloor" (kind = "group") {
+        def Scope "Housewares" (kind = "subcomponent") {
+        }
+    }
+}
+```
+
+The authored `group` kind is ignored with respect to model hierarchy because its parent isn't a `group` or `assembly`. With minimal trade-offs, this proposal argues that just as incorrect usage can be discarded by composition, _correct `group` usage can be propagated by composition_.
```
spiffmon, line 154: Technically, it is _population_ that is ignoring these errant `kind` tags, not composition.

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,342 @@
...
+        def Scope "Housewears" (kind = "subcomponent") {
+        }
+    }
+}
+```
+
+The authored `group` kind is ignored with respect to model hierarchy because its parent isn't a `group` or `assembly`. With minimal trade-offs, this proposal argues that just as incorrect usage can be discarded by composition, _correct `group` usage can be propagated by composition_.
+
+More simply-- **untagged children of groups and assemblies are automatically groups**.
+
```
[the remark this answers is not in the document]
spiffmon: "GroundFloor" becomes an ordinary, non-model, kind-less prim, as are most prims under a `component`

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,342 @@
...
+                }
+            }
+        }
+    }
+}
+```
+In this example, because `NewYorkCity` is an `assembly` (and therefore a `group`), its descendant `Water`, `Tunnels`, `Bridges`, and `Boroughs` scopes are automatically `group` prims. Propagation would end at the `BrooklynBridge`, `LincolnTunnel`, and `DepartmentStore` `component` models.
+
+### Performance When Model Hierarchy is Unused
+It's worth noting that the pseudo-root is automatically a `group`. Propagation will need to be suppressed in that case with an additional check to see if the parent is the
```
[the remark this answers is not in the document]
spiffmon: Not just the default prim but **all** root prims.  But you're right that in (our, at least) common practice, only the prim that will become the default prim would have any kind tagging in the correct case - other root prims are often `over` prims containing structure that we reference into the asset at various points.

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,342 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+```
+In this example, because `NewYorkCity` is an `assembly` (and therefore a `group`), its descendant `Water`, `Tunnels`, `Bridges`, and `Boroughs` scopes are automatically `group` prims. Propagation would end at the `BrooklynBridge`, `LincolnTunnel`, and `DepartmentStore` `component` models.
+
```
spiffmon, line 193: Before getting to discussion about performance, it would be very helpful to first give a modified example of above where there *is* some unpruned component-less path(s), and discuss whether, in terms of `IsGroup()`/`IsModel()` queries, what the prims along that path would be considered to be.  

Of particular concern to me is: as we are doing our current single population pass through the prims on the stage, is it still the case that we only compute a prim's "effective model kind" (to be cached in primFlags) **once** , or is there a need/desire to revisit ancestor prims after we reach a "model hierarchically unpruned leaf prim"?

spiffmon (commented):
Thanks for the performance-testing, @nvmkuruc !  Left one key and several ancillary questions and a couple small corrections.

proposals/selfAssemblingModelHierarchy/README.md
```diff
@@ -0,0 +1,342 @@
...
+Propagation when `instanceable=True` is complicated, as a prim may have multiple parents with different model hierarchy validity. There are actually [bugs](https://github.com/PixarAnimationStudios/OpenUSD/issues/2406) with this today.
+
+The OpenUSD development team proposes that prim flags are considered in concert with composition arcs when constructing prototypes to deal with this ambiguity.
+
+Until that fix is available, the best that can be done in both the current state of the world and with self propagating hierarchies is to encourage users to be intentional about model hierarchy when authoring `instanceable` to avoid violating the hierarchy's continuity.
+
+### Additional API
+To better clarify what it means to be a "model group", this proposal recommends deprecating `IsGroup` in favor of `MayContainComponentModel`.
+
+This proposal also recommends `IsModel` should be deprecated in favor of `MayContainOrIsComponentModel` and should be equivalent to (`MayContainComponentModel() || IsComponent()`).
```
spiffmon, line 299: I'm not averse to introducing these API's, but I don't think the rest of the proposal makes the existing `IsGroup()`, `IsModel()`, or `IsComponent()` any less correct/relevant (and all already **effective**, computed results rather than simply authored kind), so deprecating the existing methods feels incorrect.

Noting also that I am conflicted, and curious of others' thoughts, on whether, should we pursue this proposal, there would want to be updates to the Model Hierarchy description in the Kind developer documentation?  It talks about model hierarchy, but falls short of fully prescribing the population rules (which are out of scope for this low-level module anyways).

For sure we would want to add a User Guide for model hierarchy as part of this work!

spiffmon (commented):
Forgot one thing I wanted to note...

PR #16, opened 2023-07-31
Add spline animation proposal.

Pixneb:
### Description of Proposal

Add spline animation to USD.

### Contributing


phuchau1989 (approved):

proposals/spline-animation/README.md
```diff
@@ -0,0 +1,1980 @@
...
+      spline model in widest use.  Bezier curves are parametric, so evaluating a
+      curve at a given time may require an iterative solving process.  Pixar has
+      found the cost of the solve negligible in the context of executing complex
+      rigging, but it may not be fast enough for all applications.
+
+    - USD Anim will also support cubic Hermite curve segments.  These differ
+      from Bezier curves in that they have one fewer degree of freedom: tangents
+      have fixed lengths, and only their slopes may be specified.  In return,
+      Hermite curves are non-parametric, and can be evaluated in constant time
+      via a simple cubic polynomial.
```
drdogbiscuit, line 112: I think it'd be helpful to define this Hermite function. It's hard to conceptualise how this explicit Hermite function works alongside the parametric Cubic Bezier function, given the same knot data.

tallytalwar:
ready to land/review/iterate. (Merging soon via pixar-oss).

pixar-oss (approved):

JamesPedFoundry:
Hi,

I may have misunderstood some section, or missed some potential behaviour we could re-use.  However, I wonder whether the scope of this should be slightly broader to make sure we don't duplicate our efforts when it comes to curve data in general?   

For instance, shaders tend to have ramp or curve properties, which are a bit of a "wild-west" of attributes at the moment which this looks to be 80% of the way to defining.  This proposal seems to handle animated values and attributes well, but seems to miss the potential for shared evaluation with a generic curve property value which isn't necessarily `time` restricted?

I don't envisage it being a paricularly different, other than removing time from being the key point of reference for knots. It would also require having some way to query the attribute system for the value of the curve at a specific `float` value rather than time based specifically.

However, it does also bring forth another question of whether per-knot interpolation is/should also be supported?  I understand it's not widely used in animated values, but again for the case of shaders we have seen it being requested and used for shader curve definitions.

proposals/spline-animation/README.md
```diff
@@ -0,0 +1,1980 @@
...
+distributed in time.
+
+In existing USD content, animated values are represented with time samples.
+Time samples provide clear policy about render sampling: we sample at the time
+sample times.
+
+When rendering USD content containing splines, the situation will be different.
+We could potentially use knot times for sampling, but knot times don't
+necessarily indicate an intent to sample; they may only have been chosen in
+order to achieve a desired curve shape with sparse data.
```
epsilondust, line 961: Without an API to do this, I think we'd be forced to bake out the full curve or segments at the resolution we needed to capture motion blur samples.

proposals/spline-animation/README.md
```diff
@@ -0,0 +1,1980 @@
...
+See [Question: time snapping](#time-snapping).
+
+## Automatic Tangent Computation
+
+When splines contain automatic tangents, these are recomputed each time nearby
+knots change.
+
+During bulk editing operations, automatic tangent computation can be deferred in
+order to avoid redundant processing.  An RAII helper class called
+`TsAutomaticTangentBlock` will be provided for this purpose.
```
epsilondust, line 1668: Realtime feedback in curve editing operations over many knots is a big concern of ours, we would make use of this.

proposals/spline-animation/README.md
```diff
@@ -0,0 +1,1980 @@
...
+tangents of a knot may be specified separately, so a knot may be "pre-auto" and
+"post-manual", or vice versa.
+
+Automatic tangents are computed only when knots are first defined, and when they
+are edited.  We therefore say that automatic tangents are an _edit-time
+behavior_.
+
+The exact algorithm for computing automatic tangents has yet to be specified;
+most likely we will adopt one of the Maya auto-tangent algorithms.  Other
+algorithms are possible, and may be available in the future.  Ideally any
```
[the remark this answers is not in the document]
Pixneb: Totally understand the desire, but we're trying to avoid extensions, so that all content can be handled by all clients.  If there's no need for other clients to edit your knots, would it suffice to run your own auto-tangent algorithm on top of USD splines, without the auto-tangent behavior living inside the USD core?

Another avenue would be that if you think your auto-tangent algorithm is awesome and universally useful and you're willing to give it away, you could propose contributing it to USD.

proposals/spline-animation/README.md
Pixneb: Yes.  We're still cogitating on this detail, but for now my best guess is that, for Presto at least, we'll bake out time samples for these cases.

proposals/spline-animation/README.md
meshula: These Hermite curves will be as described in AnimX, c.f. https://github.com/Autodesk/animx/blob/5db8ee416d5fa7050357f498d4dcfaa6ff3f7738/src/internal/interpolators.h#L139

proposals/spline-animation/README.md
```diff
@@ -0,0 +1,1980 @@
...
+supports addition and scalar multiplication can be represented by a spline.
+Presto supports splines of floating-point vectors, but we have barely used them.
+
+Among other things, values and tangents of vector-valued splines can't easily be
+visualized in a 2D interface, making them less useful for artists.  These kinds
+of splines, if they are useful at all, would presumably be used for programmatic
+interpolation.
+
+Our guess for now is **no**, we should not support splines of vectors in USD
+Anim.
```
JamesPedFoundry, line 1838: Referring back to my message at the top level, a use case for this in a more generic curve data is for Color ramps in shaders.  These are well visualised and defined in many DCC's already for the described interpolation types.

Pixneb:
@JamesPedFoundry: can you give more detail on what "per-knot interpolation" means?

We do eventually want to support spline-valued attributes, but we think it will be survivable to leave them out of the initial release of USD Anim.  Shader ramps are a great use case.  Shaders are already being used in USD, and schemas have been put together to handle ramps, so it can be done.  Those schemas are ad-hoc and inconsistent, and spline-valued attributes should improve them, but we want to get spline-driven animation working first.

As to time-oriented vs generic splines: the plan of record, so far, is to start with only a time-oriented API, because we want to avoid building in any generality that isn't yet used.  We think swapping in a generic API with a time-oriented wrapper will be a reasonable option for the future, and we also think there would be no differences whatsoever in the implementation, only in the wording of the API.  However, another possibility is to introduce only the generic API, and have even animation refer to "x" instead of "time".  So far we're betting that would be too confusing, but we'd love to hear more feedback on the topic.

JamesPedFoundry:
Re: @Pixneb.

I understand the position here, I would still prefer a more stable API over time, rather than having to refactor later when the API inevitably changes.  However, I see the point where initial users of this may find the abstract names a little more confusing. I prefer the idea of starting with the time based API wrapper of an underlying generic API, such that we don't need to rewrite import/export logic further down the line. I imagine that would be more work, but might be the best of both worlds, with a simple interface for animated splines, but compatibility for future generic spline work.  This is all just one opinion though, and certainly won't block your excellent detailed plan described here.

On per-knot interpolation, of course I can explain more!  Essentially it boils down to being able to define the interpolation of each segment between with different interpolation methods. This may also be useful in animation, but is particularly useful in shading where you might want a gradual Bezier curve, followed by a linear interpolation to provide sharp line to signify sharp edges on the meshes (as one example).

It is likely much further down the requirements list and is likely something that could be added later if requested enough, and I understand it isn't supported by all renderers. 
An image below of a single curve that uses both linear and Bezier interpolation.
image

(Its a shame there isn't thread capability on GitHub comments. Apologies that I should have made the comment in the source to enable this, lesson learnt for the future!).

Thank you again for making this proposal, it looks like it'll be a great addition to USD going forward, and a lot of points well thought out.

Pixneb:
@JamesPedFoundry: yes, top-level (non-)threads are terrible!  See https://github.com/github/roadmap/issues/552

Certainly we intend to support per-segment interpolation methods.  I've been working for so long with splines that do this that I may have neglected to state it explicitly.  When I do my next round of doc polish, I'll make sure that's covered.  Thank you for the pretty picture, it warms my spliney heart.

We are hoping that, even if we should introduce a non-time-oriented spline class later, the API will not change.  Having decided to use a time-oriented API for animated attribute values, we will stick with that forever.  The hypothetical future change would be that the **implementation** would switch to a non-time-oriented one, and then the time-oriented version would become just a header-only set of aliases for the same classes, methods, and parameters, just with different names.  Both APIs would then be available, both doing the same thing, but with the signpost of time-oriented names when used for time-oriented purposes.

dgovil:
Just forwarding an internally sourced comment that I'd been slow to post here

> Splines of Vectors: I'd like to see this, but I think allowing things like translation to be specified as 3 separate floats solves a big driver for those. Another use-case I see is driving color attributes in shaders. Would be ok to have limited tangent controls here.
> This would allow types like vector3f and float3 to be driven by a curve whose knots are of those types

meshula:
@dgovil is the motivation behind a desire for splines of vectors a correlated interpolation of some sort? examples ~

- color attribute - Hue change minimal, or constant brightness, interpolation?
- kinematic interpolation according to the Frenet-Sennet formulae

If so, would such a need alternatively be met by something like OpenExec and a constraint system?

dgovil:
@meshula 

> Essentially, the thought was that if you had a vector3(like transform, color) that you wouldn't have to serialize them as multiple curves, but instead could essentially have the knots defined in the same type (so a curve of float3)

Anyway just wanted to forward that feedback along since the question of Splines of vectors was posited in the proposal. 
I can see the value in it, but I can also understand that it likely complicates things too, and API can probably sufficiently paper over things.

Pixneb:
@dgovil Yes, splines of vectors is a borderline case: there are some potential use cases, so we can't say a hard no, but it's also rarely used (at least at Pixar), adds complexity, and is hard for users to reason about and use.  We are confident, having implemented this already at Pixar, that it would fit into USD if we decide in the future that it's needed.  We're going to try a first version without it, and see what people come up with.  Thanks for the use case - colors are an interesting one, both in terms of how they're conventionally expressed, and in terms of how the interpolation will look, since our eyes are so sensitive to tiny color differences.

dgovil:
Thanks, Deneb! Yep that all sounds good. Thanks again for taking on this effort :-)

meshula:
Hi all, the proposal merged for easy access from the proposals front page! Like all the proposals not yet landed as part of OpenUSD, it is still open for comments. Suggestions that cover the whole proposal may be put forward via a fork and a pull request. More likely, you may have questions or suggestions about individual points; for these please open an Issue and use a permalink to reference portions of the specification you wish to comment on.

PR #18, opened 2023-08-25
Stage Level Namespace Editing

crevilla:
### Description of Proposal

This is the proposal for stage level namespace editing in USD.

### Contributing

- [X ] I agree to and accept the [Supplemental Terms](https://graphics.pixar.com/usd/release/contributing_supplemental.html).


meshula:
Rendered version here: https://github.com/crevilla/OpenUSD-proposals/tree/PROPOSALS/proposals/namespace_editing

proposals/namespace_editing/README.md
```diff
@@ -0,0 +1,480 @@
...
+
+#### Change Notifications
+
+Because namespace operations performed through the UsdNamespaceEditor API have a clear intent, we will be able to add the information about whether a UsdObject has been deleted, moved or renamed to UsdNotice::ObjectsChanged when edits are made through this API. In addition to the existing ObjectsChanged queries of `GetResyncedPaths()` and `GetChangedInfoOnlyPaths()`, we will add queries along the lines of `GetRenamedPaths()`, `GetMovedPaths()`, and `GetDeletedPaths()` to return what objects have been specifically edit using the UsdNamespaceEditor. We'll also provide the equivalent functions of `RenamedObject(UsdObject)`, `MovedObject(UsdObject)`, and `DeletedObject(UsdObject)` for querying a specific object as to whether it was namespace edited. Clients can make use of this new notice information to respond to changes a more efficient manner for namespace changes they know how to handle.
```
spiffmon, line 49: > Clients can make use of this new notice information to respond to changes a more efficient manner for namespace changes they know how to handle.

Does this imply that we will still be including ResyncedPaths for all moved/renamed objects in the same Notice, as well as info/resynced paths for all the paths that needed to be "fixed up"?  If so, it seems tricky for clients that **do** handle the new, smart changes to avoid doing all the expensive updates.. but if **not**, it seems like clients who are not in-the-know would be completely missing out?

proposals/namespace_editing/README.md
```diff
@@ -0,0 +1,480 @@
...
+
+Another option for batch editing would be to implement a BatchEditingBlock RAII structure that, when opened, would cause all edits performed via UsdNamespaceEditor to be queued up until the block is closed/deleted at which point the edits would be processed and performed. This has the advantage that we wouldn’t need an entire structure that parallels the same API as UsdNamespaceEditor just for batch editing. There's a potential downside in that we would have to process the effects of the queued edits to answer any of the “can perform edit” queries made during an open batch, but this may end up not being a big enough issue to avoid going this route. Another potential issue would result if the user wanted to change the authoring settings or edit target in the middle of an open batch. We could work around this by disallowing the changing of authoring settings or edit targets when a batch is open or by just using the state of the settings when the batch is closed.
```
spiffmon, line 57: > We could work around this by disallowing the changing of authoring settings or edit targets when a batch is open or by just using the state of the settings when the batch is closed

Could we not also just have each UsdNamespaceEditor method snap and store the current UsdEditTarget at the time it is called, rather than storing just one to associate with the whole Editor/RAII-batcher ?

proposals/namespace_editing/README.md
```diff
@@ -0,0 +1,480 @@
...
+As mentioned earlier, one major consideration when handling namespace editing in USD is that we have ways in which scene description can refer to or target namespace paths of other objects. Any scene description that targets another path may become invalid if a namespace edit were to change or remove the path of the targeted object. Thus, we will provide the ability to automatically author the necessary edits to fix up paths to namespace objects that have been moved or deleted through the UsdNamespaceEditor API.
+
+The types of paths in scene description that we will need to be able to fix up are:
+* Relationship targets
+* Attribute connections
+* Path expressions for pattern based collections
+* Composition arcs (references, payloads, inherits, specializes)
+* Overs in higher composition strength layer stacks
+* Layer metadata for default prim (which can be consumed by references and payloads)
+* Possibly internal path reliant structures (like a stage's current edit target which may have a path map function)
```
spiffmon, line 169: We may want to note that UsdCollectionAPI::MembershipQuery objects retained by clients may need to be discarded in response to namespace edits, as well.

proposals/namespace_editing/README.md
```diff
@@ -0,0 +1,480 @@
...
+
+The paths that need to be fixed up after an edit will commonly come from the layers that comprise the UsdStage that is being edited itself. But it's also possible that the edits are being performed on layers that are part of the composition of other open USD stages (for instance via references or sublayers) and we would want those stages to be fixed up as well in response the namespace edits. Because of this we will provide the ability to **explicitly designate the list of open stages on which to perform namespace target fixups** in the UsdNamespaceEditor API. Any stages specified in this list will be check for any paths that need to be fixed up and will have those edits authored to the appropriate layers as well. 
+
+The fixup behaviors we implement in USD will be modeled off of what fixups we already do in Csd. These behaviors are best explained via examples. Here we’ll define three layers root.usda, b\_ref.usda, and c\_ref.usda. root.usda defines a prim “A” which references a prim “B” in b\_ref.usda which in turn references a prim “C” in c\_ref.usda. In each layer we define connections, relationships, overs, and references to a prim that we’ll rename.
```
spiffmon, line 173: in Csd -> in Presto

proposals/namespace_editing/README.md
```diff
@@ -0,0 +1,480 @@
...
+) {
+…
+}
+```
+
+This is the only fixup necessary since the new path /XXXX is mapped across the reference in the same way as the path /C was before. It’s worth noting that if the prim B’s reference had instead been written to use the default prim instead of the explicit path to /C (a la `references = @./c\_ref.usda@`) there would be no fixups necessary in b\_ref.usda at all as the defaultPrim metadata in c\_ref.usda was fixed up to use the new prim name "XXXX".  
+
+Another thing to note is that the edit in this example can only be accomplished via a performing the edit on StageC as there’s no way to create a single edit target for StageB or StageA that can map both the old and new paths to /C and /XXXX respectively in the target layer c\_ref.usda.
+
+#### Example 2:
```
spiffmon, line 307: Really nice example of how and why this stuff gets so tricky!

proposals/namespace_editing/README.md
```diff
@@ -0,0 +1,480 @@
...
+```  
+UsdPrim aPrim = stage.GetPrimAtPath(“/A”)
+UsdPrimHandle childHandle = stageGetPrimHandleAtPath(“/A/Child”)
+```
+
+Both aPrim and childHandle would start valid. If you use the namespace editing API to rename /A to /B, then aPrim will no longer be valid since UsdPrim doesn’t track identity. But childHandle will now hold the prim object that refers to /B/Child as UsdPrimHandle will have its shared identity updated. The same would be true if we had instead deleted /A and then performed some sort of “undo” that recreated /A and /A/Child again. aPrim would still be invalid even though a prim at /A exists again, but childHandle would become valid again and hold a UsdPrim representing the recreated /A/Child.
+
+Risks, Issues and Caveats
+=========================
+
```
spiffmon, line 476: One other thing I've been thinking about... a UsdNamespaceEditor will need to do alot of work that can be cached to build up the "backwards dependencies" for relationships and connections.  If a DCC wants to provide for fast, incremental namespace edits, do we envision allowing it to retain a single UsdNamespaceEditor per-stage, which listens to general editing notices so that it can update its dependency caches (when rel/con targets change) rather than needing to fully regenerate them for each new batch of edits?

spiffmon (commented):
Great proposal, @crevilla - thank you!

proposals/namespace_editing/README.md
crevilla: Yes, I intend that ResyncedPaths will include all the moved/renamed objects precisely because it's still correct that these paths have been resynced. And the existence of downstream clients that are already looking at ResyncedPaths is an additional compelling reason as to why we'd want to keep it this way.
While it may be trickier for clients that can optimize based on rename/move to filter those out, the provided functions should be able to provide that filtering. We could additionally add an optional filter to GetResyncedPaths that would allow clients to filter out any combination of namespace edited paths if that proves to be useful.

proposals/namespace_editing/README.md
crevilla: That is a potential option, but I'm also concerned about the potential complexity in processing the effects of multiple edits spanning across many layers stacks that could have namespace dependencies on each other (and not to mention the extra complexity of properly testing it). From an implementation standpoint, I would definitely start with only allowing batch edits to involve a single edit target before trying to expand to multiple edit targets.

proposals/namespace_editing/README.md
crevilla: That's another good example; I'll add it to that bullet.

proposals/namespace_editing/README.md
crevilla: I don't necessarily envision the UsdNamespaceEditors being created per-stage as none of the API requires a persistent stage; i.e. the all the functions you would call provide the stage you're editing either explicitly (as UsdStage parameter) or implicitly (via a UsdObject parameter). But I do expect that UsdNamespaceEditor object can be persistent and would hold on to authoring settings and such. So, it's completely feasible in the implementation for it to cache dependencies that are outside of what is already being handled by PcpCaches for any stages that are affected by the edits performed.

proposals/namespace_editing/README.md
spiffmon: Thanks, @crevilla ! I reckon we can do this unambiguously for now … where it might get tricky is when we have general batch editing in usd, and then distinguishing the two cases “moved prim a and changed a property on it” from “moved prim a and switched a variant on it”.  
Both would have a resync for a, but in only one case can it be “ignored”

lbiasco-sie:
This looks great, thanks @crevilla! I have two general questions: First, are you able to give a rough estimate on how long this work might take to complete? Especially with finishing the relocates feature it seems like no small feat, so I won't hold my breath.

Second, has there been any consideration on how to manage namespace edits for stages/layers that aren't open? It seems beyond the scope of this system and is probably better left for smart use of an asset resolver, but I'm sadly coming from a games ecosystem where resolving versions is more restricted, so I'm curious if you had any thoughts on the problem.

proposals/namespace_editing/README.md
```diff
@@ -0,0 +1,480 @@
...
+#### Change Notifications
+
+Because namespace operations performed through the UsdNamespaceEditor API have a clear intent, we will be able to add the information about whether a UsdObject has been deleted, moved or renamed to UsdNotice::ObjectsChanged when edits are made through this API. In addition to the existing ObjectsChanged queries of `GetResyncedPaths()` and `GetChangedInfoOnlyPaths()`, we will add queries along the lines of `GetRenamedPaths()`, `GetMovedPaths()`, and `GetDeletedPaths()` to return what objects have been specifically edited using the UsdNamespaceEditor. We'll also provide the equivalent functions of `RenamedObject(UsdObject)`, `MovedObject(UsdObject)`, and `DeletedObject(UsdObject)` for querying a specific object as to whether it was namespace edited. Clients can make use of this new notice information to respond to changes a more efficient manner for namespace changes they know how to handle.
```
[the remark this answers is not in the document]
crevilla: Yes, that's an excellent observation! I agree with making them all `MovedObject` and `GetMovedPaths` and letting clients parse if there's more they want to glean from the paths as to how the move occurred. I'll updated this in the doc. Thanks!

melevine, line 49: Just thinking about the overall use of thee change notifications, and this may be outside the scope of the proposal, but have you also considered `GetAddedPaths()`.  In some circumstances this can remove the onus of keeping a stage prim path cache on the client side.  It may not be the same kind of namespace edit explicitly handled by the `UsdNamespaceEditor`, but feels like it might be in the same ballpark.

crevilla: It's conceivable to add `GetAddedPaths` but that's definitely outside the scope of this proposal. One advantage of having all our proposed namespace editing API in a specific new class is that we can justify tying the *new* notification about deleted and moved USD objects directly to the use of this *new* API without having to promise that we'd parse Sdf layer changes to determine delete or move intention. After typing this I'm realizing that it may make sense to have these notification functions be more explicit in that intent, i.e. `GetNamespaceEditDeletedPaths` and `GetNamespaceEditMovedPaths` or something like that.

The addition of `GetAddedPaths` would be something that should be filed as a separate request.

marktucker:
I'm having waking nightmares just thinking about the difficulties of doing this, so first let me commend your bravery! :)

A question: I don't think you specifically mention instance proxies, but I was wondering what you intend to do in the case of instance proxies as source or destination targets for deletes/renames/moves? For most normal USD edits you actually can effectively override instance proxies by adding an inherit arc to the instanceable prim. And this approach could work for the "soft delete" case you describe (and maybe for the "relocate arc" as well?). Is this an approach you've considered to allow these operations?

And I just wanted to raise a flag about "soft deletes"... Any time you "delete" a prim by deactivating it, you essentially leave a little land mine in your scene graph. If you try to create a "new" prim at or below a deactivated location, the new prim disappears into the ether. If you are smart enough to know that you need to reactivate your ancestors, then the "deleted" stuff comes back. Obviously this problem isn't unique to this feature, but I fear it will pop up a lot. A user renames A to B (which leaves a soft-deleted A lying in wait). Then the user says "no, never mind, rename B back to A" - sorry, not allowed... But if I then move B to C, and try to move C back to B, no problem! Because B could be truly deleted.

crevilla:
> I'm having waking nightmares just thinking about the difficulties of doing this, so first let me commend your bravery! :)
> 
> A question: I don't think you specifically mention instance proxies, but I was wondering what you intend to do in the case of instance proxies as source or destination targets for deletes/renames/moves? For most normal USD edits you actually can effectively override instance proxies by adding an inherit arc to the instanceable prim. And this approach could work for the "soft delete" case you describe (and maybe for the "relocate arc" as well?). Is this an approach you've considered to allow these operations?
> 
> And I just wanted to raise a flag about "soft deletes"... Any time you "delete" a prim by deactivating it, you essentially leave a little land mine in your scene graph. If you try to create a "new" prim at or below a deactivated location, the new prim disappears into the ether. If you are smart enough to know that you need to reactivate your ancestors, then the "deleted" stuff comes back. Obviously this problem isn't unique to this feature, but I fear it will pop up a lot. A user renames A to B (which leaves a soft-deleted A lying in wait). Then the user says "no, never mind, rename B back to A" - sorry, not allowed... But if I then move B to C, and try to move C back to B, no problem! Because B could be truly deleted.

Sorry for the delayed response but I can address both your questions.
As far as instance proxies we don't intend to have namespace editing de-instance any instance prims and we're going to bonk any edits that involve modifications to individual instance proxies. While adding an inherit to an instanceable prim is a way of decoupling an instance prim from its current prototype, it's not the only or even necessarily standard way of accomplishing this operation and this API doesn't feel like the appropriate place to handle how we de-instance a prim.

As far as "soft deletes", we are aware that the namespace location is now "dead" and you can't move a prim to the dead location even though there isn't a namespace prim there. This ends up being necessary because you can't truly move a different prim to the dead location without it picking up the opinions from composition that necessitated the soft-delete in the first. We actually feel that the fact that you CAN still create a new prim under a deactivated location is an issue and we have a task where hope to update the DefinePrim, OverridePrim, and CreateClassPrim methods to not allow creation of prims under deactivate prims or instance proxies.
Additionally, with the addtion of relocates, edits that required them will leave behind dead namespace locations (just like deactivation) for the original prim path. The only prim that can be moved to the dead relocated from location would be the prim that was relocated from that namespace path in the first place.

marktucker:
> As far as "soft deletes", .... with the addtion of relocates, edits that required them will leave behind dead namespace locations (just like deactivation) for the original prim path. The only prim that can be moved to the dead relocated from location would be the prim that was relocated from that namespace path in the first place.

Thank you for the responses, and particularly for this clarification. I was wondering how "relocates" would behave in this situation.

pixar-oss (approved):
(DY as pixar-oss) Approved

PR #19, opened 2023-09-13
Semantic Schema API proposal

dennis-lynch-nv:
### API to store semantic labels on prims

A simple multiple-apply schema proposal to store semantic data labels on USD prims, for use in AI/ML or Computer Vision applications.

### Supporting Materials

None

### Contributing

- [X ] I agree to and accept the [Supplemental Terms](https://graphics.pixar.com/usd/release/contributing_supplemental.html).

### DIFF
--- a/proposals/semantic_schema/README.md
+++ b/proposals/semantic_schema/README.md
@@ -1,6 +1,6 @@
 # Semantic data in USD
 
-Copyright &copy; 2022-2023, NVIDIA Corporation, version 1.0
+Copyright &copy; 2022-2024, NVIDIA Corporation, version 1.0
 
 
 Dennis Lynch  
@@ -133,13 +133,45 @@ string[] semantics:instance_name:labels = ["class:animal", "class:bird", "class:
 
 The proposed method has better performance as the number of labels of the same type increase on a prim to dozens, hundreds, or thousands of labels.
 
+### Semantic Aggregation
+
+"Aggregating" semantic data is essential to workflows.  A prim could have the semantic label of `door`, but that is missing wider context.  Is the prim a door inside of a room, on the exterior of a building, part of a vehicle?
+
+A method should be provided to "aggregate" parent semantics in a consistent way that allows users to understand the semantic heirarchy.
+
+Example prim setup:
+```
+└── xform - car
+    ├── wheel
+    ├── door
+    │   ├── window(glass)
+    │   └── handle
+    └── seat
+```
+Querying the semantic heirarchy should return an appended list/set of parents
+
+example results:
+```
+Aggregation query on handle: ['handle', 'door', 'car']
+Aggregation query on seat: ['seat', 'car']
+```
+
+### Proposed Methods
+
+The `primvarsAPI` has provided some inspiration for possible methods that would be used for semantics:
+
+```
+GetSemanticLabels() - Returns list of `labels` on prim
+GetSemanticInheritedLabels() - returns an aggregated ordered set of `labels` from the prim and its parent prims
+```
+
 ### Tokens vs Strings
 
 Consideration was taken for the datatype being tokens or strings.  Considering that semantic labels should be more read-intensive rarely (if ever) written/updated, tokens are thought to be the better representation.
 
 Some performance testing was done, and in some cases StringArray performed better than TokenArray in Python, the difference was negligible and could be attributed to the overhead of converting tokens to native Python strings.
 
-Tokes might also facilitate future expansion use of `allowedTokens` for contraining semantic labels, but that will not be a part of this proposal at this time.
+Tokens might also facilitate future expansion use of `allowedTokens` for contraining semantic labels, but that will not be a part of this proposal at this time.
 
 ## Out-of-Scope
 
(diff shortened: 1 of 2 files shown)


proposals/semantic_schema/README.md
```diff
@@ -0,0 +1,154 @@
...
+The proposed method has better performance as the number of labels of the same type increase on a prim to dozens, hundreds, or thousands of labels.
+
+### Tokens vs Strings
+
+Consideration was taken for the datatype being tokens or strings.  Considering that semantic labels should be more read-intensive rarely (if ever) written/updated, tokens are thought to be the better representation.
+
+Some performance testing was done, and in some cases StringArray performed better than TokenArray in Python, the difference was negligible and could be attributed to the overhead of converting tokens to native Python strings.
+
+Tokes might also facilitate future expansion use of `allowedTokens` for contraining semantic labels, but that will not be a part of this proposal at this time.
+
```
spiffmon, line 143: @dennis-lynch-nv , I think I recall from the ASWF discussion that a key aspect of this schema is a namespace inheritance behavior that, within a particular multi-apply "instance", the resolved value of `semantics:myConcept:labels` is the **aggregation as a list** (or is it a unique set?) of all the labels of all ancestors of the prim being queried.  That is also one of the key selling points for me as to why we want to formalize a schema here.  If this is accurate, can you please add a section that motivates and describes the behavior?

Thank you!

dennis-lynch-nv:
Hi @kevthecoder , 

Web semantics (specifically [OWL](https://www.w3.org/OWL/) and RDF) was something I looked-at when crafting this.

Overall, the conclusion was that OWL/RDF was very powerful, but also more complex than the current proposal.  Since there are many datasets that are simpler `key:value` pairs

 But, as you note), RDF can still be mapped-onto this schema if someone wishes and I think that your example is fairly spot-on with how to accomplish this.

kevthecoder:
Thanks for reviewing @dennis-lynch-nv!

proposals/semantic_schema/README.md
dennis-lynch-nv: @spiffmon - We've discussed this some internally and see your point.  We should probably do a unique set (list of Tokens?) upon traversal/collapse of the semantic hierarchy.

```
└── xform - car
    ├── wheel
    ├── door
    │   ├── window(glass)
    │   └── handle
    └── seat
```
- Query the door prim, get `semantics:myConcept:labels = ['door']`
- Query door prim for "Parents" or "Ancestors":   `semantics:myConcept:labels = ['door', 'car']`
- Query handle prim for "Parents" or "Ancestors":   `semantics:myConcept:labels = ['handle', 'door', 'car']`
- Query for "Children" or "Descendents" from door prim:  `semantics:myConcept:labels = ['door', 'handle']`
- Only semantics of the same "type"/instance name (ex: `myConcept`) are combined.  If the car prim had `semantics:otherConcept:labels = ['Incredibile']` it would not be added to the list of `semantics:myConcept:labels` when queried from the `myConcept` instance on the door or handle

What do you think?

As far as nomenclature- I wasn't sure if we would _want_ to keep the same terms that `Usd` already uses for prims (`GetChildren()`, `GetChild()`, `GetParent()`, etc.) or would we want to use different terminology to avoid confusion about Prim hierarchy traversal vs semantic hierarchy traversal

dennis-lynch-nv: "Ancestors"-only is fine, I was only including "Descendents" for reasons of symmetry, but there are not a lot/any known reasons to regularly get semantic descendants, so if it is also a performance issue, then we can leave that out.

meshula: I'm unclear if the patch visible above is the resolution of this conversation, or not?

dennis-lynch-nv: Maybe @spiffmon could comment if this is sufficient or he has something else in mind?

proposals/semantic_schema/README.md
```diff
@@ -133,13 +133,45 @@ string[] semantics:instance_name:labels = ["class:animal", "class:bird", "class:
...
+Example prim setup:
+```
+└── xform - car
+    ├── wheel
+    ├── door
+    │   ├── window(glass)
+    │   └── handle
+    └── seat
+```
+Querying the semantic heirarchy should return an appended list/set of parents
```
spiffmon: to make it more blindingly explicit, maybe something like:

... should return a uniquified list of labels, where ordering is determined by ancestral distance from the prim being queried, and the _first_ time a label is encountered in the ancestors-walk determines its position in the list.

proposals/semantic_schema/README.md
spiffmon: Added one request for clarification, but otherwise the additions look good to me!

meshula (approved):
I think we are good to merge, with Draft status. @dsyu-pixar

dsyu-pixar:
@dennis-lynch-nv apologies, would it be possible to git-sign the commits?

dennis-lynch-nv:
@dsyu-pixar - Commits have been signed

PR #24, opened 2023-12-15
Autodesk: Add text proposal

erikaharrison-adsk:
### Description of Proposal

This is [a proposal](https://github.com/autodesk-forks/USD-proposals/blob/adsk/feature/text/proposals/text/README.md) on supporting the text primitive within OpenUSD.

### Supporting Materials

N/A

### Contributing


erikaharrison-adsk:
Update: Moved textMetricsUnit from TextStyle, ParagraphStyle, ColumnStyle to SimpleText and MarkupText. Whether the unit of a metrics is world unit, pixel or publishing point should be a property of a text primitive. All the styles bound to the primitive should have the same metrics unit.

dgovil:
Thanks for putting up this PR, Erika! I'm going to confer with some folks internally to get some feedback going.
Some early feedback was:

1. Would it be possible to include some usda samples of what the data would look like?
2. Is possible to include a comparison to our [Preliminary Text schema](https://developer.apple.com/documentation/arkit/arkit_in_ios/usdz_schemas_for_ar/preliminary_text?language=objc) just for completeness. It looks like yours is much more feature full, but it would be helpful for some third party folks who'd been using our schema to envision how they'd have to upgrade their data, specifically if there are any parameters that aren't accounted for.
3. Is MarkupText the same formatting as used by Markdown? Is there a reason not to go in on HTML as the default rendering, as it is more common for UI level text in various frameworks such as Qt (and of course the web)

proposals/text/README.md
```diff
@@ -96,6 +96,7 @@ The primvars:displaycolor and the primvars:displayopacity will be the color and
...
 - primvars:backgroundColor. The color of the background. By default there is no background color. (Currently this not supported in our implementation. Will be implemented in the future.)
 - primvars:backgroundOpacity. The opacity of the background. By default there is no background opacity. (Currently this not supported in our implementation. Will be implemented in the future.)
+- textMetricsUnit. The unit for textHeight and other text metrics. It is a string token, which could be "pixel", "publishingPoint" or "worldUnit". By default it is "worldUnit". If the text primitive doesn't have billboard, only "worldUnit" is valid. If the unit is set to other value, we will still use "worldUnit" unit. The unit of the textHeight will be the same as the unit of the world space. If the text primitive has billboard, that it always faces to the camera and the size will not change, only the value "pixel" and "publishingPoint" are effective. If metrics unit is "pixel", the unit will be pixel on the screen. If metrics unit is "publishingPoint", the unit will be desktop publishing point or 1/72 inch on the screen. If the unit is set to "worldUnit", we will use "pixel" as the value.
```
meshula, line 99: Is this a correct reading?

- textMetricsUnit. The unit for textHeight and other text metrics. It is a string token, which could be "pixel", "publishingPoint" or "worldUnit". By default it is "worldUnit".

- publishingPoint and pixel are only valid if the text primitive's billboard property is true

- text height unit is always world space unit, irrespective of this setting. Therefore textMetric is for positioning only. In the case of billboards, this may result in overlapping lines of text depending on camera position.

- billboarded text primitives always face the camera and are a fixed size, so textMetricsUnit must not be set to worldUnit for billboarded text.

- pixels are pixels in the framebuffer.

- publishingPoint are 1/72 of an inch on a screen's physical display. Achieving this requires a renderer to know the physical dimensions of a monitor.

- if textMetricsUnit is set to "worldUnit", "pixel" is used regardless, so the "worldUnit" setting has no application.

Is this correct? If not, can we rewrite the section to make it more clear?

proposals/text/README.md
```diff
@@ -165,6 +165,7 @@ The primvars:displaycolor and the primvars:displayopacity will be the default co
...
 - primvars:backgroundOpacity. The opacity of the background. By default there is no background opacity. The background is completely transparent. It can be overridden by the specified background color in the markup. (Currently this not supported in our implementation. Will be implemented in the future.)
+- textMetricsUnit. The unit for textHeight and other text metrics. It is a string token, which could be "pixel", "publishingPoint" or "worldUnit". By default it is "worldUnit". If the text primitive doesn't have billboard, only "worldUnit" is valid. If the unit is set to other value, we will still use "worldUnit" unit. The unit of the textHeight will be the same as the unit of the world space. If the text primitive has billboard, that it always faces to the camera and the size will not change, only the value "pixel" and "publishingPoint" are effective. If metrics unit is "pixel", the unit will be pixel on the screen. If metrics unit is "publishingPoint", the unit will be desktop publishing point or 1/72 inch on the screen. If the unit is set to "worldUnit", we will use "pixel" as the value.
```
meshula, line 168: same comment as above. I realize this text is pre-existing, but on close read, I'm finding it difficult to understand, unfortunately.

meshula (changes requested):
Could we get a rewrite of the text for the affected paragraph please? The last bit that "worldUnit" always means "pixels" really threw into doubt my understanding of what is trying to be said here.

PierreWang:
> Thanks for putting up this PR, Erika! I'm going to confer with some folks internally to get some feedback going. Some early feedback was:
> 
> 1. Would it be possible to include some usda samples of what the data would look like?
> 2. Is possible to include a comparison to our [Preliminary Text schema](https://developer.apple.com/documentation/arkit/arkit_in_ios/usdz_schemas_for_ar/preliminary_text?language=objc) just for completeness. It looks like yours is much more feature full, but it would be helpful for some third party folks who'd been using our schema to envision how they'd have to upgrade their data, specifically if there are any parameters that aren't accounted for.
> 3. Is MarkupText the same formatting as used by Markdown? Is there a reason not to have HTML as one of the rendering methods, as it is more common for some UI level text in various frameworks such as Qt (and of course the web). That said, SwiftUI does support Markdown formatting so markdown syntax would naturally fit there as well
> 4. Similarly, for styling, has it been considered to use a CSS string? One worry is that we'd have to reimplement a lot of pre-thought out styling here whereas CSS (as complex as it is) has a lot of that legwork done. This would be also similar to how Qt defers more advanced styling to stylesheets while still supporting some parameters to be set directly.

Thank you for your comments. 
The sample usda files are included with this comment. 
[SampleUSDA.zip](https://github.com/PixarAnimationStudios/OpenUSD-proposals/files/14020718/SampleUSDA.zip)
Here is a rough compare with the Preliminary Text schema. And I can add the compare to the proposal.
Compare
We use MTEXT as the default support because it is widely used in our Autodesk product. 
We didn't consider to use CSS string for styling. Indeed our implementation was a porting from a traditional text system we already used for years.

proposals/text/README.md
PierreWang: > Is this a correct reading?
> 
> * textMetricsUnit. The unit for textHeight and other text metrics. It is a string token, which could be "pixel", "publishingPoint" or "worldUnit". By default it is "worldUnit".
> * publishingPoint and pixel are only valid if the text primitive's billboard property is true
> * text height unit is always world space unit, irrespective of this setting. Therefore textMetric is for positioning only. In the case of billboards, this may result in overlapping lines of text depending on camera position.
> * billboarded text primitives always face the camera and are a fixed size, so textMetricsUnit must not be set to worldUnit for billboarded text.
> * pixels are pixels in the framebuffer.
> * publishingPoint are 1/72 of an inch on a screen's physical display. Achieving this requires a renderer to know the physical dimensions of a monitor.
> * if textMetricsUnit is set to "worldUnit", "pixel" is used regardless, so the "worldUnit" setting has no application.
> 
> Is this correct? If not, can we rewrite the section to make it more clear?

"If the text primitive doesn't have billboard, only "worldUnit" is valid. If the unit is set to other value, we will still use "worldUnit" unit. The unit of the textHeight will be the same as the unit of the world space." This should be a separate paragraph. The unit of text Height is the unit of world space, when the text primitive doesn't have billboard. 

Yes, I think I need to rewrite the section so that it could be more clear..

erikaharrison-adsk:
Update pushed. Please let us know if this addresses things.

meshula (approved):
this reads more clearly, thanks

PR #25, opened 2023-12-19
Added photogrammetry use case

splidje:
### Description of Proposal

Recording that camera image planes would also be useful for photogrammetry workflows.

### Contributing


meshula (approved):
Good use case

PR #20, opened 2023-11-07
Autodesk: Add LineStyle proposal

erikaharrison-adsk:
### Description of Proposal

This is [a proposal](https://github.com/autodesk-forks/USD-proposals/blob/adsk/feature/LineStyle/proposals/LineStyle/README.md) on supporting line styles within OpenUSD. 

### Supporting Materials

N/A

### Contributing

- [X] I agree to and accept the [Supplemental Terms](https://graphics.pixar.com/usd/release/contributing_supplemental.html).

PR #27, opened 2024-01-18
Autodesk: LineStyle Proposal - CurveStyle Update

erikaharrison-adsk:
### Description of Proposal

Updated the LineStyle proposal based on prior feedback: https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/20 

Emphasize that we only support sketch-like line style, not 3D-like line style. We will not directly modify the basisCurves schema. Instead, a CurveStyle schema will be introduced. It is an applicable schema. The CurveStyle schema itself will contain the surface shader that will be used for style.

### Supporting Materials

N/A

### Contributing


meshula (approved):
Thanks for the update, reviewers feel free to leave comments here, or open new PRs or issues.

PR #31, opened 2024-02-09
Marked status of UTF-8 identifiers as implemented 24.03

erslavin:
### Contributing


meshula (approved):

meshula:
@erslavin Would you mind signing the commit and force pushing please? It seems like a bother, but the reason is to avoid triggering the nag-bot in future merges and integrations.

PR #36, opened 2024-03-04
Bi-Directional Transcoding of Invalid Identifiers

miguelh-nvidia:
### Description of Proposal

`TfMakeValidIdentifier`, was used in OpenUSD to convert any identifier into a valid identifier. However, it creates a 
non-bidirectional relationship, for example, something like `カーテンウォール` would be transformed into `________________`. 

The objective of this proposal is to provide an alternative to `TfMakeValidIdentifier` that can take any identifier 
(potentially with invalid characters) and transform it into a OpenUSD valid identifier.

### Contributing


miguelh-nvidia:
Pending internal review.

PR #37, opened 2024-03-07
Bi-Directional Transcoding of Invalid Identifiers

miguelh-nvidia:
### Description of Proposal

`TfMakeValidIdentifier`, was used in OpenUSD to convert any identifier into a valid identifier. However, it creates a 
non-bidirectional relationship, for example, something like `カーテンウォール` would be transformed into `________________`. 

The objective of this proposal is to provide an alternative to `TfMakeValidIdentifier` that can take any identifier 
(potentially with invalid characters) and transform it into a OpenUSD valid identifier.

### Contributing

### DIFF
--- a/proposals/transcoding_invalid_identifiers/README.md
+++ b/proposals/transcoding_invalid_identifiers/README.md
@@ -87,7 +87,7 @@ Disadvantages:
 - **Querying**, unfortunately encoding the search term and doing character comparison will not work as this is not a 
 byte-aligned encoding.This will require all paths to be decoded as they are traversed.
 - **Prefix**, to give a hint of transcoding we add a prefix `tn__`, in similar fashion to `xn--` in Punycode. 
-For short identifiers this may be big overhead.
+For short identifiers this may represent a big overhead.
 
 # Proposed API
 
--- a/proposals/transcoding_invalid_identifiers/README.md
+++ b/proposals/transcoding_invalid_identifiers/README.md
@@ -86,6 +86,8 @@ This is improved over other encoding methods, i.e. base62, where every case is o
 Disadvantages:
 - **Querying**, unfortunately encoding the search term and doing character comparison will not work as this is not a 
 byte-aligned encoding.This will require all paths to be decoded as they are traversed.
+- **Prefix**, to give a hint of transcoding we add a prefix `tn__`, in similar fashion to `xn--` in Punycode. 
+For short identifiers this may be big overhead.
 
 # Proposed API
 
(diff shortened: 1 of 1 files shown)


marktucker:
SideFX has been using a customized version of punycode to encode/decode between USD attribute names and Houdini parameter names for a very long time, and it has been quite a successful strategy. I also proposed something like this as a possible alternative to adding true UTF-8 support to USD. So I'm very much onboard with this overall proposal, and can vouch for its usefulness.

Just a few specific comments about the suggested APIs... In many cases there seems to be a typo where you are using "Boostring" instead of "Bootstring". But I would actually suggest eliminate the word "Bootstring" (or "Boostring") from the APIs completely. If this is  going to be the Sdf standard for encoding/decoding identifiers, I shouldn't have to know or care that it's using an algorithm called "Bootstring".

I'm also not sure I see the value in returning a std::optional. Wouldn't it be just as easy to return std::string() to indicate an error case? When I saw the API returning std::optional my first thought (and fear, because I think it would be a bad idea) was that the "optional"-ness was going to be used to indicate whether or not the input string had to be modified. Returning and empty std::string would also make these APIs more consistent with the signatures and behavior of TfMakeValidIdentifier.

nvmkuruc:
Thanks for the thoughts, Mark! It's helpful to know you've had success using Bootstring encoding!

> But I would actually suggest eliminate the word "Bootstring" from the APIs completely. If this is going to be the Sdf standard for encoding/decoding identifiers, I shouldn't have to know or care that it's using an algorithm called "Bootstring".

I've been an advocate for giving the encoding algorithm a distinguishing name.  My thinking is that there are still probably needs for "make valid identifier functions" that aren't bidirectional.  For example, replacing symbols and whitespace with `_` may be preferable for some users and contexts, and they're willing to give up bidirectionality.  I don't think we're attached to `SdfBootstringEncodeAsciiIdentifier` though if there's better names.

nvmkuruc:
Just wanted to note the potential need for transcoding in Hydra when mapping primvar to GLSL.

meshula:
I had to look up punystring.

> both Bootstring and Punycode serve similar purposes of encoding non-ASCII characters into ASCII-compatible formats for use in domain names, they differ in their specific applications, character support, and efficiency. Bootstring is tailored for IDNA applications, supports a broader range of Unicode characters, and can sometimes produce shorter encoded strings. Punycode, on the other hand, is more widely used for general domain name encoding, has a narrower character support, and tends to produce longer encoded strings.

Personally, I am leaning to not requiring the name of the algorithm, maybe you got my brain off on the wrong footing with the Freudian slip-ish "boostring". We'll never finish de-boosting!!

I don't believe that a preference to replace space with underbar is a sufficient argument to have the name boostring :P hang around for ever. boot string doesn't replace space with anything since the space character is already an ascii-compatible character. To me it would make sense to separate the concerns instead of making a matrix of conversions.

ie. step 1, bootstring the string. "boot string" stays "boot string". Step 2. make it identifier compatible. "boot_string". If you want to be able to make it round trippable, I think you need a second layer of escaping. so that round tripping first removes the compatibles and second removes  the boot stringing.

so "boot string" becomes "boot__string" for example. but hey, the peoples cry, two underscores is the ugly! Well, what's more important? being able to reconstruct the original string? or that the encoding is perfectly nice? You cannot have it both ways. 

Another option is that you invent a modified boot string, that takes into account that ascii characters also need escaping, so that space becomes __0x20 or somesuch.

meshula:
PS if we are down to bikeshedding the name, shall we merge this PR in Draft state?

meshula (approved):
Looks ready to merge in Draft state.

meshula:
@miguelh-nvidia Are you able to sign the commits and force push? If no, I'll merge as is. Otherwise I'd prefer to attempt to satisfy the automation bot.

miguelh-nvidia:
@meshula ready!, thanks.

PR #42, opened 2024-03-26
Add visualization schema proposal

nvmkuruc:
### Description of Proposal

The proposes introducing a shared schema domain to collect and standardize behaviors of primitives designed for nondiegetic visualization.

### Contributing

### DIFF
--- a/proposals/vprim/README.md
+++ b/proposals/vprim/README.md
@@ -1,0 +1,121 @@
+# Visualizing the Unseeable
+
+## Overview
+OpenUSD models both fantastical and real world scenes with physically modeled geometry and materials. There's an increasing need to render nondiegetic visualizations that do not participate in physically based rendering and simulation. These may be user defined labels or curves describing a path through a scene. They may be curves representing vector fields (or even captured light paths!). The traditional method of tagging `Gprim`s with `guide` purpose has not been sufficient in part because users are looking to visualize in non-physical ways (like screen space widths). These nondiegetic visualizations require different descriptions than those the current `UsdGeom` domain currently specifies.
+
+### Nondiegetic visualizations and `guide`s
+We use "nondiegetic" as a handle for describing objects intended to be a part of any primary visualizations of the scene but aren't physically present. If `Gprims` are diegetic like dialogue on audio track, we're concerned with how to represent the visualization equivalent of the orchestral score. Nondigetic effects cannot be removed without fundamentally changing the viewing experience but hold a different set of constraints and idealized representations than their diegetic cousins.
+
+We view nondiegetic visualizations as orthogonal to the `guide` `purpose` attribute. In our audio track analogy, `guide` serves the role of a DVD commentary, an overlay adding additional context and visual information that aren't a part of the primary viewing experience and may be optionally experienced.
+
+## Background
+OpenUSD was developed as path tracing and physically based rendering were maturing in the visual effects production pipeline. As such, scenes are modeled as physical 3-D geometry. `Points` are spheres and `BasisCurves` are tubes (or user-oriented ribbons) that can be ray traced from any direction. They were decidedly not the screen oriented disks and ribbons previously used in rasterizers.
+
+Users are currently seeking a way to visualize data through schemas added to OpenUSD. A user may want describe a "red line" to draw attention to a particular object or a "green curve" to represent available paths towards exits during building design.
+
+### Schema Slicing
+One instance of these new schemas is `CurveStyleAPI`, proposed as an API schema for the existing `Curves` schemas. The API schema makes three transformative changes. It transforms the meaning of the `widths` attribute to be a screen space size (`constant` or `uniform` only), changes the appearence of joints and end caps, and resticts the materials allowed to be bound to a restricted set of "styling" materials.
+
+We believe this creates a situation we're calling "schema slicing" (to adapt the term "object slicing" from C++). In that language, object slicing occurs when the properties and behavior of a derived type are "sliced" off when interfacing with an object via a higher level type. Using `CurveStyleAPI` to transform the curves into screen space defined primitives similarly creates the potential for slicing. A user interacting with the prim using the `BasisCurves` schema and not the full prim definition will have effectively "sliced" off the reinterpretation of widths and other modifications to the underlying definition. Utlimately, "slicing" complicates interchange and standardization.
+
+### Disaggregating `hermite` out of `BasisCurves`
+As a historical analogue, the `hermite` basis was previously a part of `BasisCurves` but was left unimplemented in Storm and many renderers. We learned the primary usage was to describe paths and not as a renderable representation, ultimately seperating it out into its own schema. This change made `BasisCurves` simpler for renderers to fully support and let `HermiteCurves` stand on its own with a decoupled `points` and `tangents` representation and without the need to formulate support for the different `wrap` modes of `BasisCurves`.
+
+We similarly wonder if all variations of the current set of `Curves` are needed to describe screen space visualizations and when adding support, how renderers and other consumers know which to prioritize to establish compatability. A new axis of variation that transform the existing geometric schemas complicates interchange and compatability of the `UsdGeom` domain. We suggest new schema domain with a new set of schemas may be the best place to drive development of nondiegetic primitives separate from their diegetic `UsdGeom` cousins.
+
+## Proposal
+Introduce a class of visualization primitives that are defined in a 3-D scene, but whose imaging definition is allowed to include nondiegetic screen effects and with a relaxed definition of boundability. Let's call them `Vprims` (visualization primitives). Boundability as a schema partioning mechanism is already employed in `UsdLux` which classifies light schemas as either boundable (ie. `RectLight` and `SphereLight`) or non-boundable (`DomeLight`).
+
+Let's use points as a motivating example. Consider an OpenUSD file that contains a mesh reconstruction of data from a point cloud scan. A user might benefit from visualizing the points on top of the mesh to aid in debugging some errors in the reconstruction. The user might choose the `Points` schema to encode positions. And they might be able to hide the source point cloud from rendering contexts using `purpose` tagging and/or variant sets. But they will be unable to make the points all the same size in camera space in a specification compliant renderer. Preview optimizations that might give a user their desired effect like those found in Storm exist primarily for performance reasons and may not persist across renderers, sessions, and future releases.
+
+A visualization focused specification of points could look something like this--
+
+```
+def VizPoints "PointCloud" {
+    point3d[] points = [...]
+    float[] pixels = [...]
+    color3f[] primvars:displayColor = [(1, 0, 0)]
+}
+```
+
+`VizPoints` could be rendered in camera but may also be composited against a depth buffer from a physically based renderer.
+
+### Styling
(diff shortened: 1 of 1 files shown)


proposals/vprim/README.md
```diff
@@ -0,0 +1,121 @@
...
+* When defining screen space sizes, what is the best choice of units? Is it pixels or something else?
+* How should scaling be treated for screen space objects?
+
+## Future Work
+
+### Hiding
+As currently described, visualization primitives should be strictly hidden via depth. However, one could imagine customizing hiding for these prims. Some visualizations might want to hide based on a particular reference point for the entire string of text. Some visualizations might want to express priority with respect to other objects in the scene. Hiding is another axis where the specification may eventually end up diverging from the `Gprim` space.
+
+### Breaking the Fourth Wall
+Some care needs to be taken to establish best practices and rules for visibility of screen-constrained primitives. While we generally would expect nondiegetic primitives to not participate in physically based lighting, consider screen constrained "context spheres" as reference objects. They "break the fourth wall" and capture light and reflections. It should be possible to specify this behavior even when using nondiegetic transformations.
```
spiffmon, line 114: This one somewhat throws my thoughts about the proposal into disarray... it doesn't submit to the same clean "let a rasterizer handle it in a post-pass", and it also goes against the final section below, as a fully formed material specification is required to capture light and reflections...

spiffmon (commented):
There's alot I like here, and would like to see some rendering-folk perspectives on it.

Also:
* Would Viz prims be minimally UsdGeomXformable?  Or they just refer/constrain to Xformable or BOundable prims?
* Formatting-wise, if you use text-fill or add linebreaks yourself within paragraphs, it allows more precise commenting in the gitHub UI.

proposals/vprim/README.md
nvmkuruc: Look for some clarifications in a revision shortly.  My hope would be that we could establish some set of default behavior for objects that works for most visualizations and consider if additional schemas or configurations are needed to bridge the gaps.  (It's possible the set of effects that can be described using schemas without a material network ends up being too narrow.)

nvmkuruc:
> There's alot I like here, and would like to see some rendering-folk perspectives on it.
> 
> Also:
> 
> * Would Viz prims be minimally UsdGeomXformable?  Or they just refer/constrain to Xformable or BOundable prims?
> * Formatting-wise, if you use text-fill or add linebreaks yourself within paragraphs, it allows more precise commenting in the gitHub UI.

I think `Vprims` would generally be `Xformable` yes.

proposals/vprim/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+### API Schema and Grouping Adapters
+This proposal frames screen space points (and paths) visualization primitives as fundamentally different from their `Gprim` counterparts, warranting their own first class `Vprim` typed schemas. Does this proposal set the precedent that other `Gprim` types like `Cube` or `Cylinder` require visualization equivalents if they wanted to be adapted into the "visualization" domain? We'd suggest no and offer two solutions that have prior art in OpenUSD.
+
+If a primitive retains its underlying definition, `UsdLux` already provides a solution to the problem of adapting `Gprim`s into another domain. Rather than a `MeshLight`, the `Mesh` `Gprim` is adapted via an API schema. `Gprim`s could be similarly "adapted" into the visualization domain. To avoid "schema slicing" an adaptation must not change the underlying definition or require their own distinct `purpose` or `visibility` tagging. Consideration must be also given to if or how these schemas affect interactive picking. A non-slicing visualization effect might include sillhoutte outlines, glows, or overlays applied to a `Gprim` to style or highlight its importance.
```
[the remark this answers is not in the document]
tcauchois: I guess there's a philosophical question; practically, we'll implement overlay and silhouette as image-plane operations because we don't want them to participate in rendering and after rendering is done image-plane is the best way to express things.  Any kind of image operation can reasonably expect to get a certain amount of data, e.g. color, depth, id, etc if needed.  I think silhouette can be done as an image operation as an edge-detect convolution on primId?  And overlay can of course be done trivially as a primId match.

Internally we do some more complicated things where we render additional 3d geometry with custom compositing, e.g. drawing rotation widgets on top as transparent if they fail the depth test and as opaque if they pass the depth test.  I don't know if we want to try to model that stuff here.  In games that's something that can be modelled by materials, but since that kind of thing isn't physically based it's not captured by UsdShade.  It's certainly useful for those kinds of prims to identify that they don't participate in light transport, at least, and I could imagine having hydra support for rendering 3d prims into the scene after the primary render (e.g. by running Viz prims through Storm and comping them in).

proposals/vprim/README.md
tcauchois: It does seem like there are a few entangled concepts here:
1.) Objects whose position or size are dependent on the camera.
2.) Objects that don't participate in light transport.
3.) Possibly redundant with 1 & 2, but stylizations done in the image plane.  Objects in categories 1 and 2 can be placed in the physical scene, if not in USD then at least in the renderer, and I suppose a silhouette *could* be constructed in the 3d scene, but that wouldn't be my choice for implementation.

If we handle concerns 1 & 2 separately, having screen-constrained primitives that participate in light transport is easy.

nvmkuruc: Agreed-- I'm working on a revision as we speak that tries to formalize some of the ideas and came to the conclusion that it's better to keep some of these concerns separate.

proposals/vprim/README.md
```diff
@@ -0,0 +1,300 @@
...
+* `VizOverlayAPI`: Add a nondiegetic color overlay on top of an
+existing `Gprim`
+* `VizSilhoutteAPI`: Add a nondiegetic edge on top of an existing
+`Gprim`
+* `VizScreenAnchor`: Nondiegetic screen anchoring transformation
+
+### Schema Slicing
+One important consideration when introducing schemas in this domain is
+to whether or not they are new drawables (`Vprim`s) or styles on top
+of existing drawables (`Gprim` + `VizStyleAPI`).
```
tcauchois, line 186: I understand the concerns about schema slicing, but I also think that this approach will require (in the long term) a ton of duplicate schema definitions for everything in the Gprim hierarchy, which I don't love.  At the hydra level, I'd plan to implement these with the same usdImaging and hydra code and set appropriate flags for xform computation and lighting participation based on Vprim/Gprim; I'd strongly resist much duplication there because the maintenance burden would be significant.

proposals/vprim/README.md
```diff
@@ -0,0 +1,300 @@
...
+grey sphere. The viewer will see the red transparently overlayed, but
+the sphere is still a grey sphere from the perspective of light
+transport (reflected and refracted light).
+
+Consideration must be also given to if or how these schemas affect
+interactive picking. A non-slicing visualization effect might include
+sillhoutte outlines or overlays applied to a `Gprim` to style or
+highlight its importance.
+
+### Example: `VizScreenAnchor`
```
tcauchois, line 130: A note on implementation thoughts; I left a bunch of comments on the Autodesk billboards proposal along these lines:

My preferred way to do screen/camera anchoring is to augment the hydra "xform" definition to say that a transform is rooted in view space.  This matches what GlfSimpleLight does for camera lights, and is consistent with the other flag we have in the hydra "xform" definition: resetXformStack, which can be thought of as saying that a transform is rooted at the origin of world space.  The idea is that these xforms would flow to the renderer, and Storm (for example) could turn a view-relative transform into a world-relative transform in the shader, meaning that we don't need to recompute the transforms for each viewer in a multi-viewer app.

If we want to separate position anchoring, rotation anchoring, and scale anchoring, it's a bit more complicated but you can say (for example) that a matrix is translate-anchored to the camera and just compensate the translation as above.

I think it is cool and useful to let transforms below one of these camera-anchoring operations compose, since the composed transform would be (implied inverse view xform) * parent xform * child xform, and the child composed transform just inherits the flag indicating the implied inverse view xform.

This is all consistent with VizScreenAnchor being a scoping mechanism, and would potentially be with an xformOp as well.

Any kind of screen space positioning, scaling, styling is perfectly xformable and boundable in the renderer, of course, but absent a GfFrustum for context the UsdGeomXformCache/UsdGeomBboxCache become meaningless at some point in the tree, so for non-hydra consumers I suppose it's important to flag when that becomes the case.  I think if XformCache/BboxCache had an option to take a GfFrustum, it could replicate the compensation that hydra is doing, and that would be a useful utility to provide.

proposals/vprim/README.md
nvmkuruc: The intent is definitely not to duplicate the `Gprim` hierarchy.  To me, the main grey area (with the current hierarchy) is how to handle screen dependent points and curves.  With curves, there are already multiple schema definitions and they're already partitioned between those intended for production rendering (`BasisCurves`) and annotation and rigging purposes (`HermiteCurves` and `NurbsCurves`).

However, that doesn't necessarily mean that `Vprim` is necessarily the right answer to that problem.  Without introducing a higher level `Vprim` class, you could disambiguate curves that are "wires" and "tubes/ribbons".  Wires without widths would be "intrinsically boundable" by their points and a "styling API" could be used to give wires nondiegetic effects like pixel widths.  This would (perhaps narrowly) avoid a "slice" (like interpreting a single pixel width as a meter width) because consumers unaware of or unable to process styles would simply process the curve as an infinitely thin wire.  I'd argue that the disambiguation between "wires" and "non-wires" would be best done at the schema level, but presence of `widths` and `normals` could be used as well.  (I suspect that if "wire-ness" is not specified at the schema level, validators will flag curves authored without `widths` as errors which is why I'd lean that way.)

proposals/vprim/README.md
meshula: using my few functional brain cells to parse "nondiegetic break in a single diegetic space";;;;; 

"nondiegetic break in a single diegetic space" 

means

 "a thing on the stage that is not part of the world represented by the stage". If that's correct, could your amendment parenthetically add the explanation for people who need more sleep?

proposals/vprim/README.md
nvmkuruc: I think consumers like hydra that can guarantee they have a single frustum have a lot of flexibility with how this gets implemented.

In my latest updates, I'm suggesting that transforms and bounds might want to be evaluated with respect to anchoring points-- to most prims that would be the pseudo root and nothing would change, but you'd be able to identify the anchoring point and use whatever extrinsic context was needed (resolution, frustum, etc.).

proposals/vprim/README.md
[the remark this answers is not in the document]
nvmkuruc: Everything you said makes sense to me Tom-- I'd expect that most styles would be implemented as image shaders that way.

I suspect there's probably going to eventually be a need for shaders to support more advanced effects, but I also suspect that there's a lot of concepts that benefit from a simplified initial specification.

proposals/vprim/README.md
nvmkuruc: Thanks Nick.  I'll try to clarify that statement more in my revisions.  The original idea of this document assumed that all nondiegetic effects were nonphysical and therefore should be excluded from lighting.  Talking to Dhruv and others, more narrow use cases emerged where it makes sense to have physically based materials and geometry constrained to view orientation.

So the revision for "anchoring" will introduce no constraints on lighting which aligns with Tom's observation that a few assumptions and usages were entangled.

PR #45, opened 2024-04-04
Revise Use of Layer Metadata in USD

spiffmon:
### Revise Use of Layer Metadata in USD

As OpenUSD's use has grown since its initial release, especially in disciplines 
outside its initial target of animation and VFX, some elements of its design have
been exercized and leveraged beyond Pixar's initial vision for them.  One, in 
particular, has not conceptually scaled well to the complexity of scenes people now
often build.

This proposal re-examines USD's use of *layer metadata*, first explaining the various
ways in which it is currently used in USD, then noting where and why it has been 
problematic.  Finally it will provide new guidelines for when to encode a concept in 
layer metadata, and propose that several core concepts currently expressed as layer
metadata be migrated to **applied schemas**.

### Supporting Materials

Rendered proposal text [here](https://github.com/spiffmon/OpenUSD-proposals/blob/main/proposals/revise_use_of_layer_metadata/README.md).

### Contributing

- [ X] I agree to and accept the [Supplemental Terms](https://graphics.pixar.com/usd/release/contributing_supplemental.html).


dgovil:
I generally like this proposal. I had a question: does everything in these APISchemas also automatically become a valid layer metadata key? I assume we still allow people to set the "global layer" hint as we do today, but taking Physics and ColorSpace, would they also allow for globally setting the keys too?

spiffmon:
@dgovil , I would argue to deprecate (at least strongly discourage authoring with warnings, while keeping backwards compatibility with a long tail) the existing global layer hints (except, as stated, for the color config data), and nothing that we introduce as properties in API schemas would be reflected as layer metadata.  I know we can think of it as a "fallback", but it also becomes a ODR issue that will create confusion (I just set the global layer data, but it's not having any effect, #$@#$!!)

dgovil:
Would the expectation then be that every single prim needs to author these? Or would it be inheritable through the hierarchy such that only the top level of any hierarchy needs it?

meshula (approved):

PR #46, opened 2024-04-05
Guidelines for new proposals

dgovil:
### Description of Proposal

This is meta proposal that provides suggestions for putting up new proposals, such that they'll be easier to digest and provide feedback on.

Things like what information a proposal could include, structuring guidance and recommendations to include schema snippets.

This isn't pointed at any specific proposals, but it is feedback we've received internally when trying to collect feedback in the past. It's something I believe would make it easier to get more feedback and higher quality feedback.

[Link to Rendered Proposal](https://github.com/dgovil/USD-proposals/blob/ProposalGuidance/README.md#guidance-for-writing-proposals)

### Contributing

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,38 @@ When a proposal has taken enough shape that it warrants detailed feedback and it
     1. The PR may include links to supporting materials that could not be included with the proposal files, such as white papers. Add links to the "Supporting Materials" section of the PR body.
     2. Please mention and link any issues and PRs in [OpenUSD](https://github.com/PixarAnimationStudios/OpenUSD) or [OpenUSD-proposals](https://github.com/PixarAnimationStudios/OpenUSD-proposals) that are related to the new proposal. A label will be created to link relevant discussions together. 
 
+### Guidance for Writing Proposals
+
+When writing and submitting a proposal, we make the following suggestions for receiving the best feedback.
+
+1. Include a link to the rendered proposal. This can be as simple as linking to the markdown file in your source branch.
+   
+   This makes it significantly easier for someone to read the document in a well formatted way.
+2. Your PR and proposal should include, and ideally start with, a short summary of what your change would achieve.
+
+   A possible structure could be:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -33,6 +33,42 @@ When a proposal has taken enough shape that it warrants detailed feedback and it
...
+   5. **Details** about your proposal, such as why you are making certain choices
+   6. **Risks** that you anticipate (if any)
+   7. **Alternate solutions** that you have considered (if any), and why you didn't go with them
+   8. Any **excluded topics**, that you have left out and why. These may be things you want to handle separately in the
+      future for example.
+
+3. It is highly recommended that any submitted proposal include text examples of what your proposal would look like
+   in `.usda` syntax.
+4. It's recommended that long sentences be split over multiple lines in your Markdown file.
+   This allows more granular feedback, and easier viewing of the files. A recommended line length is 120 characters.
```
meshula, line 59:
```suggestion
   This allows more granular feedback, and easier viewing of the files.
```
I suggest we don't recommend line lengths. Doesn't seem worth opening a debate about "holy" line lengths which specifying 120 would seem to beg.

meshula (changes requested):
One tiny request then it's good to land.

meshula:
Thanks for this, it'll definitely help people structure proposals and understand what makes a proposal useful.

README.md
dgovil: Removed!

meshula (approved):

meshula:
@dgovil Any idea what's going on with the Unverified tag on your commit? You didn't signoff the commit, so that perhaps could have caused the Unverified, but clicking on it lists reasonable looking certs but says "The signing certificate or its chain could not be verified."

dgovil:
I'm not sure. I just tried re-signing but like you say: it sees that it's signed , it just doesn't see the signature as verified.
It's probably some mismatch between internal and external signing. I'll look into it.

Issue #49, opened 2024-04-14
Language tags for text proposal


meshula:
@PierreWang wrote:

>Q: Language attribute may be required, so that the implementation will know what language the script is.
> A: We can decide the script from the unicode of the character. Because the font contains a limit set of charsets, the charset of the character may not be supported by the current font. In that case, we may display the character using a default character (such as a blank rectangle), or use font substituion to find another font which can display the character. So I think we don't need to add the language attribute.

In https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/48 @dgovil wrote:

You have various issues like [Han unification](https://en.m.wikipedia.org/wiki/Han_unification) that prevent a script being truly inferable, but outside of that it's generally not an issue.

A language consideration for the text proposal would make sense.

It is usefully separate from accessibility as well since it could be shared across multiple domains. Imagine a scenario where you have text in your scene but want to present different text per language as well.

In effect what I'd propose is that the USD file have a stage level metadata (and possibly prim level with Spiff's new proposal) to specify language. I'd also propose that this be allowed to be a purpose on attributes for when it differs from the primary language.

Something like

```
#usda 1.0
(
     language = "en_ca"
)

def foo {
     string text = "Colours are awesome"
     string text:en_us = "Colors are awesome, but the letter U is not"
     string text:fr = "La couleur est géniale"
}
```

The benefit here is that you can also specify translations in separate USD files and layer them in.

Another reason I'd advocate for language tags is that a theoretical screen reader could enunciate words differently based on the language specified.

meshula:
@dgovil 

On accessibility, would there be a need for a suffix that says "tts" (text to speech) or other affordance? Or are those concerns strictly orthogonal?

Would you imagine that USD would normatively list "fr", "en_us", "braille" etc., or would it normatively reference a master list somewhere?

meshula:
@spiffmon 

Capturing your feedback here:

> I really like the multiple language encoding idea, @dgovil . Perhaps not as germaine for labels on parts, but for anything more conversational or longer, it'll be useful to be able to intentionally provide non-auto-generated translations.

And I'd advocate skipping layer metadata entirely for the "default encoding" tag, and go directly to an API schema; in multi-user environments, you'll likely see avatars being referenced in from many different parts of the world, I'd imagine?

meshula:
Further, on alt-text:

> Q: The "alt-text" solution if we can not support the character.
[...]

PR #55, opened 2024-04-16
Multiple Language Support

dgovil:
### Description of Proposal

This proposal adds a human language local to USD so that content may be localized or provide hints for assistive technologies.

[Link to Rendered Proposal](https://github.com/dgovil/USD-proposals/blob/language/proposals/language/README.md)

### Disclaimer

This proposal isn't representative of upcoming functionality in our products. This is a future looking proposal that seemed timely to augment the proposals by others here like Autodesk.

This builds on discussion in [Issue 49](https://github.com/PixarAnimationStudios/OpenUSD-proposals/issues/49)

### Supporting Materials

An example

```
def Foo(
    prepend apiSchemas = ["LocaleAPI"]
    language = "en_US"
) {
    string text = "There's a snake in my boot"
    string text:fr_CA = "Il y a un serpent dans ma botte"
    string text:hi = "मेरे जूते में एक सांप है"
}
```
### Contributing

### DIFF
--- a/proposals/language/README.md
+++ b/proposals/language/README.md
@@ -155,6 +155,25 @@ def foo {
 
 One advantage of this system is that you can have your translations in different layer files and referenced it in.
 
+It would be recommended that at least one version of the attribute exclude the language token, so that it can 
+fallback to the inherited language and also be used as the fallback if a user asks for a language that has no matching
+languages available.
+
+#### Token Delimiter
+
+The proposal currently implicitly uses the last token as the language.
+It might however be preferable to be explicit about this by also prefixing `lang_` or `lang:` to the last token.
+
+This could look like one of the below
+
+```
+string text:lang_en_us
+string text:lang:en_us
+```
+
+This perhaps makes things longer, but does make it easier to discern that a token represents a language in a wider
+range of use cases. 
+
 #### Why not Variants?
 
 Variants are also a possible solution, however we believe that this becomes difficult for systems to work with as
@@ -183,6 +202,9 @@ Using the example above, a call to `GetLanguagePurposes(foo, "text")` would give
 - (en_US, foo:text:en_US)
 - (fr, foo:fr)
 
+In this case, the `<Fallback or Unknown>` represents that it should follow the logic 
+in the `Unspecified Language Fallback` section.
+
 I would suggest another function like
 
 ```
@@ -192,6 +214,21 @@ TfToken UsdLocaleAPI::GetFallbackLanguage(const UsdPrim& prim)
 That would return either the inherited value or a sentinel `Unknown` value when no language is specified.
 Perhaps USD could have some convenience function to do user locale lookup, but I do not think that needs to be a
 requirement.
+
+#### Language Selection Recommendations
+
+When the requested or desired language is not represented in the set of languages within the file, there are some recommendations
+on how a runtime can pick a fallback option.
+
+Following the recommendations of CLDR and BCP-47, we suggest:
+
+1. If a language is available within the set of languages, pick that attribute. e.g. `en_US` matches `text:en_US`
+2. If a language isn't available, check for a more specific version of that language.
+   e.g. `de_DE` matches `text:de_DE_u_co_phonebk`
+3. If a more specific language isn't available, then pick a less specific purpose.
+   e.g. `en_US` matches `text:en`
+4. If a less specific version isn't available, take the version without any language specified.
+   e.g. `en_US` matches `text`
 
 ## Risks
 
--- a/proposals/language/README.md
+++ b/proposals/language/README.md
@@ -1,0 +1,214 @@
+# Multiple Language Support
+
+[Discussion](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/55)
+
+## Summary
+
+We propose additions to USD to allow specifying the human language locale used so that content may be
+localized or provide hints for assistive technologies.
+
+We propose use of [BCP-47](https://www.w3.org/International/core/langtags/rfc3066bis.html) specifiers according
+to the [Unicode CLDR](https://cldr.unicode.org) specification, using underscores as the delimiter.
+
+We propose specifying the language as metadata on prims as well as a purpose on attributes.
+
+```
+def Foo(
+    prepend apiSchemas = ["LocaleAPI"]
+    language = "en_US"
+) { 
+    string text = "There's a snake in my boot"
+    string text:fr_CA = "Il y a un serpent dans ma botte"
+    string text:hi = "मेरे जूते में एक सांप है"
+}
+```
+
+## Problem Statement
+
+Today, most 3D formats assume a single unspecified language across the represented content.
+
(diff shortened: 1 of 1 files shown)


PierreWang:
I understand the usage of this language id now. The proposal looks good to me. I have one suggestion for the fallback behavior. In my opinion, if an attribute doesn't specify a language, it is a fallback for any language. Its language can be inferred by the runtime or the system locale, but we also need to allow the user know that it is a fallback.

For example, 
def foo {
     string text = "Colours are awesome"
     string text:en_us = "Colors are awesome, but the letter U is not"
     string text:fr = "La couleur est géniale"
}
The first attribute is the fallback attribute, and the language may be inferred as "en_gb". If the user want to display the text in language "ar" (Arabic), it may not find a suitable attribute. In that case, it should be able to know that the first attribute is the fallback one.
So GetLanguagePurposes may return a map which contains:
(fallback, foo:text)
(en_gb, foo:text)
(en_us, foo:text:en_us)
(fr, foo:fr)

dgovil:
Ah yeah that makes sense. I'll add in the wording accordingly.

meshula:
@PierreWang I think the comment from Dhruv  in the other thread ran along the lines of "we can't infer based on codepoints because of ambiguity".  When you say `en_gb` can be inferred, I think you mean that it can be inferred from system locale. I don't think it makes sense that the default fallback is the language of the system locale.  If I wrote my default text in England, and you read it in France, the text did not become French because you opened it in another country :) I propose that `en`  with no regional suffix should be the standard fallback, if the language attribute was not specified.

proposals/language/README.md
```diff
@@ -0,0 +1,251 @@
+# Multiple Language Support
+
+[Discussion](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/55)
+
+## Summary
+
+We propose additions to USD to allow specifying the human language locale used so that content may be
+localized or provide hints for assistive technologies.
```
cookiecrook, line 8:
```suggestion
localized to provide language and locale context for rendered text, speech synthesis, assistive technologies, or other applications.
```

proposals/language/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+We propose use of [BCP-47](https://www.w3.org/International/core/langtags/rfc3066bis.html) specifiers according
+to the [Unicode CLDR](https://cldr.unicode.org) specification, using underscores as the delimiter.
```
cookiecrook, line 11: 👍🏼

proposals/language/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+A few changes and upcoming changes to USD increase the need to specify language:
+
+1. With Unicode support in USD, it is more attractive to people in a wider range of locales.
+2. Upcoming text support feels like a natural area for representing content in different locales
+3. USD is now used as part of interactive content (games, spatial computing), where
+   localization for user playback and assistive technologies may be useful.
+
+Since there is no language specification, it is unclear for tooling and users how content should be interpreted
+when used by language aware technologies.
```
cookiecrook, line 38:
```suggestion
when used by language-aware technologies.
```

proposals/language/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Summary
+
+We propose additions to USD to allow specifying the human language locale used so that content may be
+localized or provide hints for assistive technologies.
+
+We propose use of [BCP-47](https://www.w3.org/International/core/langtags/rfc3066bis.html) specifiers according
+to the [Unicode CLDR](https://cldr.unicode.org) specification, using underscores as the delimiter.
+
+We propose specifying the language as metadata on prims as well as a purpose on attributes.
```
cookiecrook, line 13: Are any doc references needed here? E.g. https://openusd.org/dev/api/class_usd_prim.html

proposals/language/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+### Unspecified Language Fallback
+
+In the event that a language is not specified, it is recommended to specify a fallback behaviour.
+
+Our recommendation is:
+
+1. If your attribute or prim is missing a language, check the parent hierarchy for an inherited value
+2. If no language is specified, and if your runtime can infer a language, it is free to do so but does not have to.
+3. If you cannot or chose not to infer a language, assume the users current locale.
```
cookiecrook, line 105:
```suggestion
3. If you cannot or chose not to infer a language, assume the user's current locale.
```

proposals/language/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+This matches the behaviour of common assistive technologies like screen readers.
+
+### Default Metadata
+
+Most content will prescribe to one primary language, which tends to be the region of the content creator.
+To facilitate this, we encourage but do not require content authors to specify a language.
```
cookiecrook, line 112: I agree with "encourage but not require." 👍🏼 There's a lot of HTML content in the wild marked as `en-us` when it's clearly not.

proposals/language/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+For this we recommend that languages specification be a purpose on the attribute rather than having a single
+attribute language.
+
+Our recommendation is for this to be the last token in the attribute namespaces to work towards the most specific.
+
+```
+def foo {
+     string text = "Colours are awesome"
+     string text:en_us = "Colors are awesome, but the letter U is not"
```
cookiecrook, line 151: 🤣

cookiecrook (commented):
Minor editorial comments. Looks good overall. Thanks!

spiffmon:
> @PierreWang I think the comment from Dhruv in the other thread ran along the lines of "we can't infer based on codepoints because of ambiguity". When you say `en_gb` can be inferred, I think you mean that it can be inferred from system locale. I don't think it makes sense that the default fallback is the language of the system locale. If I wrote my default text in England, and you read it in France, the text did not become French because you opened it in another country :) I propose that `en` with no regional suffix should be the standard fallback, if the language attribute was not specified.

Just plussing that, in case @PierreWang meant... if the text were instead "Lorries are awesome", it's not reasonable for a DCC to infer that's `en_gb` even though it is...

spiffmon (commented):
THis looks great, @dgovil - just a couple requests for clarification.

proposals/language/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+However, per the new proposal this should move to an API schema, and we'd propose the following
+
+```
+def Foo(
+    prepend apiSchemas = ["LocaleAPI"]
+    language = "en_CA"
+) { ... }
+```
+
```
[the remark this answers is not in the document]
dgovil: Cool, I changed it to a uniform string

meshula (approved):
ready to move to draft

spiffmon:
@dgovil , @meshula , should we move this to "Hold" status in favor of [Localization with Language Catalogs](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/78), which supercedes it?

dgovil:
That sounds good to me!

PR #59, opened 2024-04-18
USD Anim update: floating-point scalars only; anti-regression.

Pixneb:
### Description of Proposal

- This change simplifies the USD Anim proposal to support only floating-point scalar splines, where before we proposed supporting all attribute types.  It removes the "series" generalization, and the QuatSeries, LerpSeries, and HeldSeries classes.  These classes were largely duplicative of time samples, and we now propose doing without them.  See the "Value Type Categories" section for a more detailed discussion.

- A much more detailed proposal is given for "regressive" splines, which go backward in time and must be forbidden.  Regressive splines are probably never created deliberately, but the Bezier math permits them, and there are many different stategies in broad use for preventing them.  The new anti-regression proposal focuses on flexible authoring-time strategies.

[Link to Rendered Proposal](https://github.com/Pixneb/OpenUSD-proposals/blob/spline-animation-update/proposals/spline-animation/README.md)

### Contributing


alankent:
Thanks for pushing this project forwards! I look forward to seeing the end result. Feedback in case useful. 

An example USDA sample would be nice to add.

I was wondering about the potential to use the same animation data in a USD Sequence schema in the future (OTIO based or otherwise). That is, allow a sequence editor have clips of animation data, and allow sliding them around (adjust start/end times), blending between tracks, etc. I know this is not part of the proposal (and I am not suggesting adding it), but I was wondering if value clips could form a basis for such an approach in the future. I saw that value clips are excluded for now. Is there a technical problem to overcome there? Or is it just "lets get V1 working, then worry about value clips".

meshula:
Note that blending animation sources is likely in the realm of OpenExec. The thought here is that animation blending varies so much per use case that it might not be formalizable in a useful way, and so would need a procedural description.

PR #62, opened 2024-04-30
Autodesk: Update Text proposal according to comments

erikaharrison-adsk:
### Description of Proposal

Update the proposal according to the comments.

We will focus on the schema in the proposal, not the implementation. So I removed the part about the Rprims, and also the plugins.

Add more detail about the SimpleText and MarkupText schemas. Add how the extents should be implemented.

Add the language tag for the textData in SimpleText and markup in MarkupText, and refer to the language proposal.

Add the "plain" tag for markup which is an alternate string when the markup can not be identified.

Remove the renderer property in SimpleText and MarkupText.

Change the name of the properties in TextStyle. Add altTypeface property which is used when the main typeface can not support some characters.
Add details about the font weight.

The TextLayout and TextLayoutAPI are merged into one schema. The new TextLayoutAPI not only contains the character directions, but also contains the direction how the lines are stacked.

Modify the name of the properties in ColumnStyle. Remove the direction and linesFlowDirection property in ColumnStyle. Instead, you can apply a TextLayoutAPI to a ColumnStyle to specify the directions.

In the part of "Some implementation details", discuss the extrusion, the missing characters, the layout for complex script and the billboard.

Finally there is a big part of four examples.

[Link to Rendered Proposal](https://github.com/autodesk-forks/USD-proposals/blob/adsk/feature/text/proposals/text/README.md)

### Supporting Materials

Previous PRs:

* https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/48
* https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/38
* https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/24

### Contributing

### DIFF
--- a/proposals/text/README.md
+++ b/proposals/text/README.md
@@ -76,11 +76,11 @@ The work proposed here aims to tackle all of these properties. Those not listed 
 There are already many different techniques which can display the text on screen. One commonly used technique is MSDF, that generate the multi-channel signed distance field for each character, and use it to reconstruct the shape of the glyph. Another technique is to get the control points of the curves which form the outline of the characters, and then render the curves on the screen. Different techniques may have different quality and performance. 
 
 # The single line single style text schema
-A new primitive SimpleText is added, which inherits from Gprim. It inherits properties from Gprim.
+A new primitive SimpleText is added, which inherits from Gprim. It inherits properties from Gprim. The primitive is for a single line single style text. "Single line" means that the baseline of the characters is straight and there is no line break. "Single style" means the appearance style for the characters are assumed to be the same. Here, we use "assume" because the user would like that the style is the same, but in the implementation, a part of the characters may not be supported so it may use an alternate style to display the characters. That is, although in schema level we use one text style for the SimpleText, on the screen some characters may still be rendered with a different style.
 
-We also have 2 schemas that the SimpleText primitive can bind. They are TextStyle and TextLayout. All the text schemas are in the UsdText domain. We have default value for TextLayout, so SimpleText primitive doesn't need to bind to TextLayout. But one SimpleText primitive must bind to a TextStyle.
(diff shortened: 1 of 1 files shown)


PierreWang:
@meshula @dgovil @tcauchois @spiffmon
This is the updated text proposal. I hope that all your suggestions are included. Please give me more comments if you have any concern. Thank you.

The biggest change is that I removed the detail about the rendering method. In original proposal we put most of the fragment shader code into getOpacity function, and the function is implemented together with a surface shader. It seems it is not the purpose of the material shader. So do you think that we should directly add a default and complete implementation in text.glslfx, and if the user would like to use a different implementation, they will modify the text.glslfx and add their own implementation?

meshula:
Thanks for the update Pierre! 

> It seems it is not the purpose of the material shader. So do you think that we should directly add a default and complete implementation in text.glslfx, and if the user would like to use a different implementation, they will modify the text.glslfx and add their own implementation?

Yes, this is what I personally would hope for. The reason I want this is that I personally would prefer to have exact control of text shading and rendition, with opacity and color being computed jointly in order that I may perfectly control aliasing.

meshula (approved):
This is a substantial improvement! Thank you for your attention to detail.

I'm going to leave this open a little longer before merging in order that others may comment.

PierreWang:
Hi @meshula . I found an issue with the proposal. In the future we are going to implement MSDF for the rendering of text. We need to create a texture and the texture will be used as a shader resource. So the text primitive still need to bind a material. Without the material binding, I don't know how to use a texture as a shader resource in USD.

PierreWang:
We have created two prs for the schemas of SimpleText and MarkupText.
The schema for SimpleText: https://github.com/PixarAnimationStudios/OpenUSD/pull/3258
The schema for MarkupText: https://github.com/PixarAnimationStudios/OpenUSD/pull/3259

PR #63, opened 2024-05-22
Extended unicode identifiers propossal

miguelh-nvidia:
### Description of Proposal

UTF-8 identifiers adhere to [Unicode XID](https://unicode.org/reports/tr31/) and as such, some identifiers are still invalid, in particular identifiers with leading digits and medial hyphens (i.e. `-`).

This type of identifiers are consistently used in different areas like manufacturing and it is important to have a direct one-to-one mapping between the virtual representation and the real world.

In this proposal we address their inclusion in OpenUSD.

### Contributing

### DIFF
--- a/proposals/extended_unicode_identifiers/README.md
+++ b/proposals/extended_unicode_identifiers/README.md
@@ -1,0 +1,299 @@
+# Extended unicode identifiers
+
+Copyright &copy; 2024, NVIDIA Corporation, version 1.0
+
+Miguel Hernandez  
+Matthew Kuruc
+
+# Contents
+
+- [Introduction](#introduction)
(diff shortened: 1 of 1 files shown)


asluk:
let's a add a link to the rendered proposal to the PR description - https://github.com/NVIDIA-Omniverse/USD-proposals/tree/extended_unicode_identifiers/proposals/extended_unicode_identifiers 👍🏼

dgovil:
This would also be greatly useful to us, especially with regards to the language schema proposal.
I'd also advocate for this + the variant set metadata proposals because they have a material impact on our AOUSD spec. I think they'd be very good to get into the first version of the spec so that we don't need to rev the spec so soon after it would be out.

PR #64, opened 2024-05-22
MaterialX Versioning in USD

ld-kerley:
### Description of Proposal

This proposal records the MaterialX library version in USD, allowing use of the MaterialX upgrade functionality.

[Link to Rendered Proposal](https://github.com/ld-kerley/OpenUSD-proposals/blob/materialx_versioning_proposal/proposals/materialx-versioning/README.md)

### Supporting Materials

An example

```
def "Material" MyMaterial ( 
    prepend apiSchemas = ["MaterialXInfoAPI"]
)
{
    uniform string mtlxinfo:version = "1.39"
}
```

### Contributing

### DIFF
--- a/proposals/materialx-versioning/README.md
+++ b/proposals/materialx-versioning/README.md
@@ -15,6 +15,16 @@
 We propose adding attributes to UsdShade `Material` primitives to record the MaterialX library version they were authored
 with. This will allow for MaterialX to provide an automatic upgrade functionality to the data stored in OpenUSD.
 
+The following is an example of a `Material` primitive with the applied schema applied.
+```
+def "Material" MyMaterial ( 
+    prepend apiSchemas = ["MaterialXVersionAPI"]
+)
+{
+    uniform string materialXVersion = "1.39"
+}
+```
+
 ## Introduction
 From its inception, the MaterialX project has considered MaterialX documents to be an archival format.  What goes along 
 with this is an implied promise that opening a MaterialX file from 10+ years ago should generate a material that will 
@@ -48,19 +58,8 @@ We propose adding an applied API schema `MaterialXVersionAPI` (in the spirit of 
 ["Revise use of Layer Metadata in USD"](https://github.com/PixarAnimationStudios/OpenUSD-proposals/blob/main/proposals/revise_use_of_layer_metadata/README.md)) 
 that can record the MaterialX library version on the UsdShade `Material` primitive.  
 
-The following is an example of a `Material` primitive with the applied schema applied.
-```
-def "Material" MyMaterial ( 
-    prepend apiSchemas = ["MaterialXVersionAPI"]
-)
-{
-    uniform string materialXVersion = "1.39"
-}
-```
-
-Adding the version to the `Material` 
-primitive allows for referencing the material directly from a `.mtlx` file. i.e.
-
+Adding the version to the `Material` primitive allows for referencing the material directly from a `.mtlx` file and 
+still retaining the MaterialX library version when the scene is composed. i.e.
 ```
 def "Material" MyMaterial (
     references=@myMaterials.mtlx@</MaterialX/Materials/MyMaterial>
--- a/proposals/materialx-versioning/README.md
+++ b/proposals/materialx-versioning/README.md
@@ -19,7 +19,12 @@ with. This will allow for MaterialX to provide an automatic upgrade functionalit
 From its inception, the MaterialX project has considered MaterialX documents to be an archival format.  What goes along 
 with this is an implied promise that opening a MaterialX file from 10+ years ago should generate a material that will 
 render a qualitatively similar image.  This promise is honored by the project in the form of an automatic system that 
-upgrades data as documents are loaded to be compatible with the current MaterialX library version.
+upgrades data as documents are loaded to be compatible with the current MaterialX library.
+
+The MaterialX library is a collection of XML files that describe a standard set of node definitions, type definitions and 
+other components provided by MaterialX.  The Material library version is a `major.minor` version number (currently 1.38) 
+and allows that set of definitions to be versioned.  It is not the version of the MaterialX project itself, which also 
+includes a `patch` version number, but the MaterialX library version does track in-sync with the MaterialX project version.
 
 The MaterialX integration within OpenUSD relies upon two libraries, `UsdMtlx` and `HdMtlx`. `UsdMtlx` transcodes the 
 MaterialX document in to `UsdShade` primitives. Inside of Hydra, `HdMtlx` is used to reconstruct the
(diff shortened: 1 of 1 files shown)


ashwinbhat:
Thanks for this proposal, I haven't followed many of the discussions around this apologies in advance if this was already discussed.
I want to clarify some workflow expectations using this minimal example:
```
#usda 1.0
(
    defaultPrim = "sphere"
    upAxis = "Y"
    metersPerUnit = 0.01
)

def Scope "MaterialX" (
    references = @./usd_preview_surface_plastic.mtlx@</MaterialX>
)
{
}

def Sphere "sphere" (
    prepend apiSchemas = ["MaterialBindingAPI"]
)
{
    rel material:binding = </MaterialX/Materials/USD_Plastic>
}

```
Config 1: 
If the OpenUSD + MaterialX config is: 
-  OpenUSD is built MaterialX **1.39**
-  usd_preview_surface_plastic.mtlx (uses materialx version="**1.37**")
-  uniform string mtlxinfo:version = "**1.38**" 

Since usd_preview_surface_plastic.mtlx is authored using version 1.37. 
Is it safe to assume that this will require UsdMtlx to upgrade the referenced usd_preview_surface_plastic.mtlx to **1.38 (instead of 1.39 SDK version)** to maintain hdMtlx compatibility?

Config 2: 
Similarly, if the config is:
-  OpenUSD is built MaterialX **1.39**
-  usd_preview_surface_plastic.mtlx (uses materialx version="**1.39**")
-  uniform string mtlxinfo:version = "**1.38**" 

Would this require MaterialX to write 1.38 using 1.39 MaterialX SDK?

ashwinbhat:
> @ashwinbhat , Config 1 can't work that way, because of how UsdMtlx works. I.e., UsdMtlxFileFormat, upon reading a mtlx file, will **apply** MaterialXInfoAPI to each UsdShadeMaterial prim it creates, and the version will always be the version from the mtlx document. In your scenario, the only way the "resolved" `mtlxinfo:version` could be other than 1.37 is via an overriding stronger opinion, **authored outside/above the reference** . It will always (and only?) be the job of HdMtlx to perform any upgrades (or rather, get MaterialX to do it on the documents it reconstructs). So when doing codegen, MaterialX-via-HdMtlx will **erroneously** upgrade from 1.38 to 1.39. I think we'd want to discourage (but cannot prevent) manually overriding the generated `version` string, unless some process is actually doing the upgrading and encoding the results back into USD.
> 
> For Config 2, I think `version` is always a statement of what is there in the document, not a statement of what MaterialX inside of HdMtlx needs to write (i.e. I'd imagine HdMtlx would upgrade the 1.38 Material (again, erroneously, given its source) to 1.39). If you're asking about a hypothetical MaterialX document-writer that translates from USD documents, then I guess you could make an argument that it should use the version specified in the USD document rather than the application's version, though "upgrade to current" would likely be an export option you'd want?

@spiffmon thanks for clarifying this, I misread the usdmtlx section: 
_This version number would come directly from the .mtlx file that was being transcoded to OpenUSD._
so in my examples configs,  
for config 1: mtlxinfo:version will be "1.37" since mltx document is 1.37
for config 2: mtlxinfo:version will be "1.39" since mtlx document is 1.39

proposals/materialx-versioning/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Proposal to Version MaterialX data in OpenUSD
+### Overview
+We propose adding an applied API schema `MaterialXInfoAPI` (in the spirit of the 
+["Revise use of Layer Metadata in USD"](https://github.com/PixarAnimationStudios/OpenUSD-proposals/blob/main/proposals/revise_use_of_layer_metadata/README.md)) that can record the MaterialX library version. We keep the name of the 
+applied API schema a little more generalized to leave the door open for additional MaterialX document information to be
+added to the schema at a later date. We deliberately decide to only propose adding the MaterialX library version to the 
+schema to keep the proposal as uncontroversial as possible to add rapid adoption.
+
+The `MaterialXInfoAPI` will be applied to the UsdShade `Material` primitive. This allows for referencing the material
```
tallytalwar, line 65: If the aim is to only restrict this APIs application to UsdShadeMaterial, then we should also consider "apiSchemaCanOnlyApplyTo = UsdShadeMaterial". Refer documentation for the same here: https://openusd.org/dev/api/_usd__page__generating_schemas.html

proposals/materialx-versioning/README.md
```diff
@@ -0,0 +1,130 @@
...
+Specifically we do not propose attempting to validate that all opinions that make up a composed
+`Material` primitive were authored with the same MaterialX version, even though this composition could lead to an
+invalid MaterialX material, for a number of reasons.
+* The expense of the validation, in some circumstances, could be too high to consider.
+* It is possible that a `Material` primitive may still contain a valid MaterialX material even if composed of differing
+  MaterialX library versions.
+* Reporting the error during the evaluation of Hydra may be later in a pipeline than desired.
+* Some implementations might not even use the `HdMtlx` library to extract the MaterialX material.
+
+Instead, we propose to leverage another OpenUSD proposal, the [USD Validation Framework](https://github.com/PixarAnimationStudios/OpenUSD-proposals/tree/main/proposals/usd-validation-framework). Adding a `UsdValidator` to
```
tallytalwar, line 107: Sounds good, should work if appropriate MaterialX APIs can be used to determine these incompatibilities.

tallytalwar (commented):
This looks good to me, just a small suggestion here.

proposals/materialx-versioning/README.md
spiffmon: Plus to this... will prevent ambiguous applications of putting versions directly on Shader prims or other prims, and restrictions can be lifted later, but not the other way round.  Question, though: will NodeGraphs need to be versioned independently within a document (e.g. if they are referenced from another mtlx file, or can we assume that every mtlx document referenced onto a UsdStage is internally version-consistent?

proposals/materialx-versioning/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+### UsdMtlx Implementation changes
+
+The only change required in `UsdMtlx` would be to apply the applied API schema to the `Material` primitive, and set the
+MaterialX library version.  This version number would come directly from the `.mtlx` file that was being transcoded to 
+OpenUSD.
+
+### HdMtlx Implementation changes.
+
+The `HdMtlx` library will also need to be changed to read the MaterialX library version from the `Material` primitive,
```
spiffmon, line 89: Slight clarification: it will be the usdImaging scene adapter that processes Materials that will need to be updated to pass on the version info, which HdMtlx can then consume from the Hydra dataSource.

proposals/materialx-versioning/README.md
```diff
@@ -0,0 +1,130 @@
...
+to the default if missing. 
+
+Specifically we do not propose attempting to validate that all opinions that make up a composed
+`Material` primitive were authored with the same MaterialX version, even though this composition could lead to an
+invalid MaterialX material, for a number of reasons.
+* The expense of the validation, in some circumstances, could be too high to consider.
+* It is possible that a `Material` primitive may still contain a valid MaterialX material even if composed of differing
+  MaterialX library versions.
+* Reporting the error during the evaluation of Hydra may be later in a pipeline than desired.
+* Some implementations might not even use the `HdMtlx` library to extract the MaterialX material.
```
spiffmon, line 105: You are now able to write a UsdValidator test that can perform such a validation independently of stage-construction (UsdMtlx) or rendering.

ld-kerley:
Are there any changes you want to see here before moving forward and working on a PR?

meshula (approved):

meshula:
@ld-kerley I've moved the status to Published. Reviewers, was @tallytalwar 's comment resolved?

> If the aim is to only restrict this APIs application to UsdShadeMaterial, then we should also consider "apiSchemaCanOnlyApplyTo = UsdShadeMaterial". 

was @spiffmon 's comment resolved?

> Slight clarification: it will be the usdImaging scene adapter t

All other conversations have been marked resolved.

PR #66, opened 2024-06-05
Clarify and quantify cost of additional kind reads for "Self Assembling Model Hierarchy" proposal

nvmkuruc:
### Description of Proposal

This change updates the proposal to clarify that under most circumstances, additional kind reads are not required and quantify the cost of the additional reads when those reads may be required.
 
https://github.com/NVIDIA-Omniverse/USD-proposals/blob/selfAssemblingModelHierarchies/proposals/selfAssemblingModelHierarchy/README.md

### Contributing


meshula (approved):
Thanks for these performance notes!

PR #67, opened 2024-06-28
Add initial draft of captions extension to semantic labels proposal

nvmkuruc:
### Description of Proposal

The semantic labels schema will add support for labeling subgraphs with tokens from discrete taxonomies.

This proposes a peer schema in the `usdSemantics` domain for `semantics:captions` to capture instances of natural language descriptions.

[Link to Rendered Proposal](https://github.com/NVIDIA-Omniverse/USD-proposals/blob/captions/proposals/semantic_schema/captions.md)

### Contributing

### DIFF
--- a/proposals/semantic_schema/captions.md
+++ b/proposals/semantic_schema/captions.md
@@ -1,0 +1,98 @@
+# Semantic Captions
+
+Copyright &copy; 2024, NVIDIA Corporation, version 1.0
+
+## Goal
+Extend the `usdSemantics` domain with a schema for natural language 
+semantic descriptions of subgraphs of an OpenUSD stage.
+
+## Background
+The proposed semantic labels API provides a mechanism for tagging
+subgraphs with labels using different taxonomies.
+
+```
+def Mesh "orange" (apiSchemas = ["SemanticsLabelsAPI:tags",
+                                 "SemanticsLabelsAPI:state"]) {
+    token[] semantics:labels:tags = ["fruit", "food", "citrus"]
+    token[] semantics:labels:state = ["peeled"]
+}
+```
+
+Labels as discrete tokens makes it easy to construct merged 
+sets of ancestors (to aid `IsLabeled` queries) or generate
+segmentation images. Labels also suggest certain user interfaces and 
+query considerations (ie. drag and drop labels onto prims).
+
+## Proposal
+Some workflows are looking for natural language descriptions of
+subgraphs that can be used to describe a variety of features.
+
+```
+def Xform "dancing_robot" (
+    apiSchemas = ["SemanticsCaptionsAPI:summary",
+                  "SemanticsCaptionsAPI:skills"]
+) {
+    string semantics:captions:summary = "This is a bipedal robot made up of metal that is painted red. The robot was manufactured in 2024."
+    string semantics:captions:skills = "This robot can perform a box step, a grapevine, and the waltz."
+}
+```
+
+Captions are expected to be longer phrases, sentences, or paragraphs.
+A single prim may have multiple named instances of a caption to
+suggest its intent. Multiple instances can also allow for multiple language
+support.
+
+### Time Varying Captions
+Just as labels benefit from being able to describe state changes 
+over time, captions may also be time varying.
+
+```
+def Xform "learning_robot" (
+    apiSchemas = ["SemanticsCaptionsAPI:skills"]
+) {
+    string semantics:captions:skills.timeSamples = {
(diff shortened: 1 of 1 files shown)


dgovil:
Hey @nvmkuruc ,
We have a **very** similar proposal centered around Accessibility that I've unfortunately been slow to put up as we discuss internally. I wonder if we should discuss the overlap because I think there's applicability here to other use cases.

For example, this is what I would propose for accessibility information

```
def Mesh "Cube" (
    prepend apiSchemas = ["AccessibilityAPI"]
)
{
    string accessibility:label = "A Cube"
    string accessibility:alternate:default = "This cube is a wonderful looking cube"
    token accessibility:importance:default = "Standard"
    
    string accessibility:alternate:size = "As big as a house"
    string accessibility:importance:size = "Low"
}
```

This follows standard accessibility forms across multiple accessibility frameworks in multiple browsers and operating systems etc... where Label is the short description and alternate is a longer description should someone want that.

Much like yours, I use `namespace:<label|alternate|importance>:<optional purpose>`

I would also (in ours) encourage combining use with the [proposed language schema](https://github.com/PixarAnimationStudios/OpenUSD-proposals/tree/main/proposals/language) so that you could do things like

```
#usda 1.0
(
    language = "en_ca"
)

def Mesh "Cube" (
    prepend apiSchemas = ["AccessibilityAPI"]
)
{
    string accessibility:label = "A Cube"
    string accessibility:alternate = "This cube is a wonderful looking cube"
    token accessibility:importance = "Standard"
    
    string accessibility:label:default:lang:fr = "Un cube"
    string accessibility:alternate:default:lang:fr = "Ce cube est un cube magnifique"
    string accessibility:alternate:default:lang:fr_ca = "Ce cube est un cube magnifique canadien"
}
```

I realize that the semantic description and the accessibility description might differ, but the abject schema seems so similar, that I wonder if we shouldn't combine forces.

nvmkuruc:
@dgovil Do you happen to have an example of allowed / preferred values for the `importance` tokens?

dgovil:
Probably just low, standard, high. It's not as common a metadata but it does help prioritize tokens for a system when someone asks with natural language for a description

proposals/semantic_schema/captions.md
```diff
@@ -0,0 +1,98 @@
...
+```
+def Xform "learning_robot" (
+    apiSchemas = ["SemanticsCaptionsAPI:skills"]
+) {
+    string semantics:captions:skills.timeSamples = {
+        0 : "The robot does not know how to dance",
+        100 : "The robot is learning the box step",
+        150 : "The robot knows the box step"
+    }
+}
```
cookiecrook, line 58: As mentioned in @dgovil's [proposal](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/69#pullrequestreview-2157301430), we also discussed future consideration for time-based descriptions. Sometimes a relevant time sequence needs an "announcement" for assistive technology, too... Either tied to a transition (for example, between slide builds where the change is more important than either end state) or a time code of the overall timelines, similar to closed captions or audio descriptions. 

Potentially relevant: I'm working on a [PR for VTT to add an ATTRIBUTES block](https://github.com/w3c/webvtt/pull/523), generally to disambiguate various types of metadata, but specifically because it's a prerequisite for using VTT to define time-based [general flash data (seizure avoidance, etc.)](https://github.com/w3c/webvtt/issues/512) in this follow-up VTT issue.

cookiecrook: I'm curious if you see these `timeSamples` keys aligning with other timed-text formats like VTT.

nvmkuruc: If I understand `VTT` correctly, it specifies time code ranges while OpenUSD holds an authored value until the next authored time sample and pulls from the first or last time sample when querying out of range. To describe that format in OpenUSD, you'd likely have to do something like this--
```
string  semantics:captions:skills.timeSamples = {
   99.9999: "" # Suppress out of range queries
   100 : "This is some state." # This is valid between time codes 100 and 150.
   150.00001: "" # Suppress out of range queries
}
```

This is highly speculative, but I'm curious if there's path to building something like VTT using the time series proposal as a starting point. It's currently designed for animation splines, but might provide a path for eventually describing more complicated time based value resolution?

spiffmon: Time series have actually been removed from the design for animation splines, in favor of more simply leveraging timeSamples for all non-scalar, non-floating-point varying data.

PR #78, opened 2025-03-26
Localization with Language Catalogs

dgovil:
### Description of Proposal

This proposal is an update to the prior Language proposal, which tries to incorporate much of the feedback since the initial proposal.

It also tries to incorporate some of the use cases that have arisen since then, as well as tries to aggregate common patterns with other localization formats.

[Link to Rendered Proposal](https://github.com/dgovil/USD-proposals/blob/languageCatalogs/proposals/language/README.md)

The primary change is that localizations have been moved from the prims themselves to a separate localization catalog.
This reduces a few of the concerns folks had including:

1. Storing a lot of potentially unused attributes on a prim, and having to parse all the prim identifiers to build the translation map. These are now stored external to the prim.
2. Knowing which prims are translated. There is now a metadata flag for it
3. Translation of non attribute types shown in UIs like Variant Sets and Variants, as they can now also be translated with this configuration
4. Doubling up of data if two attributes share the same string

I believe this is a more elegant system as a result of that feedback loop.

### Contributing

### DIFF
--- a/proposals/language/README.md
+++ b/proposals/language/README.md
@@ -5,23 +5,40 @@
 ## Summary
 
 We propose additions to USD to allow specifying the human language locale used so that content may be
-localized to provide language and locale context for rendered text, speech synthesis, assistive technologies, or other applications.
+localized to provide language and locale context for rendered text, speech synthesis, assistive technologies, or other
+applications.
 
 We propose use of [BCP-47](https://www.w3.org/International/core/langtags/rfc3066bis.html) specifiers according
 to the [Unicode CLDR](https://cldr.unicode.org) specification, using underscores as the delimiter.
 
-We propose specifying the language as [metadata](https://openusd.org/release/glossary.html#usdglossary-metadata),
-or as an [attribute](https://openusd.org/release/glossary.html#attribute)
-on [prims](https://openusd.org/release/glossary.html#usdglossary-prim) as well as a purpose on attributes.
+We propose specifying the language as a binding to a language catalog, and a new metadata key to signify
+localizations are available.
 
 ```
-def Foo(
+def Xform "Sherrif" (
     prepend apiSchemas = ["LocaleAPI"]
+    localized = true
 ) { 
-    uniform string locale:langue = "en_US"
-    string text = "There's a snake in my boot"
-    string text:fr_CA = "Il y a un serpent dans ma botte"
-    string text:hi = "मेरे जूते में एक सांप है"
+    rel locale:translations = </Foo/Translations>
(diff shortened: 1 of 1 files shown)


dgovil:
Hey Spiff, thanks for the in-depth response.

Regarding the first couple questions:
1. I think the localization itself would not be time varying BUT there is the chance that the source string to be localized might be time varying. I don't have a specific use for that myself though other than perhaps accessibility data. In that case the localization would be static to the input string, but would vary along with it in time? Again, just thinking hypothetically.
2. I think long strings might come up, although they'd be rare. I'm thinking of the use of something like a New York Times supplemental piece or museum infographic that might have diegetic text in the future. I do think they'd be rare though...

**Catalog Encoding**
I think having a dictionary of `{language: {source: translation}}` would make sense, and as you mention, sparse composition would mean we'd keep the benefits of letting people do sparse additions. I always forget that dictionaries have that ability. 

 I don't have a strong preference for whether the language identifier is foremost or the string itself. I think there's probably a benefit of having the language first, so you can just keep the parts of the dictionary you need.

I do think people might want to (in the future) also apply localization to asset paths like audio and textures. I don't think we need to tackle that any time soon, but I think it's worth considering while making the catalog encoding. But I think a dictionary would scale fine there too.

**Inheritance Override/Merge of Catalogs?**
My one concern was if someone referenced in a prim with a partial catalog onto a prim halfway up the hierarchy, causing the translation to break , but I don't think that will be a common case.
I think taking your suggestion of closest ancestor with no merging makes sense for performance reasons, and someone could reference in the full catalog if they need it if they accidentally break things.

**localized Metadatum**
I don't think we need it. My thinking was that you could prevent unnecessary lookups, but I think maybe it's premature optimization on my part and we can skip it for now.

One thing that someone did mention is that HTML allows for the inverse `<p translate="no">Something Important that we shouldn't trust a robot to auto-translate</p>` (https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/translate) but I feel like that's also something we could add in the future and don't need to bake in the first version.

**Variants and VariantSets**
Oh perfect, yes, if we could get display names for variants, that is really the thing I want to translate. Our use case is that we use USDZ with variant sets on the root prim as a configurator and currently display the variant identifier with some string shenanigans. But if we could use a display name there it would let us do more interesting things and have fewer workarounds.

spiffmon:
That all sounds great, and that kind of planned/structured use of long strings is clever and sounds great.  Regarding timevarying, yes that's exactly what I was thinking, also - the translations are entirely value-based, and shouldn't care what timeCode the value was extracted from.

PR #81, opened 2025-04-29
Initial LOD proposal

meshula:
### Description of Proposal

This proposal defines an API schema for managing Level of Detail (LOD) in USD stages. The LOD API enables creators and renderers to efficiently switch between different detail levels of assets based on configurable criteria like distance, screen size, or explicit selection. This approach allows optimizing rendering performance while maintaining visual quality where needed.

This proposal is a modernization of the original Ubisoft LoD proposal which is included in the proposal as supporting documentation. Many thanks to the original authors for that work.

This proposal is preliminary, for community discussion and requirements gathering, and not expected to land in its present form. Please use this PR as a place to collate discussion and feedback. It would be great to learn of alternative approaches, modern LoD systems, how LoD works in _your_ engine, and so on.

[
[Link to Rendered Proposal](https://github.com/PixarAnimationStudios/OpenUSD-proposals/blob/main/README.md)
](https://github.com/PixarAnimationStudios/OpenUSD-proposals/tree/lod/proposals/level-of-detail)

### Contributing


meshula:
The proposal has been updated to reflect feedback to date, and now includes a definition of Heterogeneous Hierarchical Level of Detail.

> How could we do paired LOD? E.g if you have crowds, you might have mesh+skeleton+material that change together versus independently.

@dgovil The proposal has been updated to concretely cover that case.

meshula:
Proposal updated with most recent feedback taken into account.

meshula:
The schema definition has been updated to reflect most recent feedback.

dgovil:
So just some feedback items:

* I think the API names are a little too generic. ConfigurationAPI and LevelAPI aren’t clear by themselves and would preclude anyone else having something called configuration api. Which I can see people wanting that in the future for other domains too

* same with selectionMode and targetFramerate, unless that gets namespaced when applied? 

* I think we’re still missing how to tell USD not to render a specific hierarchy until it’s activated by an LOD. The example scene I think would have people rendering the city hierarchy in its entirety. 


But from a “what needs to be defined” perspective I think it makes sense.

meshula:
after you run it through genSchema, it comes out like UsdLodConfigurationAPI and UsdLodLevelAPI, but I suspect you are looking for something else when you say they aren't clear. Open to suggestions, I'm coming up dry.

These are multi-apply schemas so they become namespaced in the usual way, yes.

As to rendering a specific hierarchy, when you select a particular level, that "level" may contain a hierarchy. You could set up a sub-hierarchy within the city, the schema is intentionally hierarchical. Presumably you would set up some kind of heuristic to prune the city; the example doesn't set that up. But you might divide the city into blocks, and set up a frustum heuristic as an example so that only blocks within the six bounding planes of the frustum are traversed, as an example. There's no frustum heuristic proposed, however the mechanism is meant to make it simple enough to declare one.

dgovil:
The GenSchema name would just be the C++ API but I mean what would be serialized in the USD file. It wouldn't have the prefix applied so would preclude any other api in any other schema being called ConfigurationAPI. 

For pruning, I feel like that's something we should be solving along side this proposal. Leaving it up to heuristics and how a renderer or endpoint may load the scene feels like it would be fragile and lead to lots of divergence in behaviour

spiffmon:
The schema declaration can be configured to preserve just the "Lod" part of the prefix for the schema name/identifier, while still allowing the C++ classes to have the full "UsdLod" prefix, [as we do for the schemas in UsdRender](https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/usd/usdRender/schema.usda#L131-L140).

dgovil:
Yeah I think that would be preferable. That way it's LODConfigurationAPI which feels much nicer in the file.

proposals/level-of-detail/example.usda
```diff
@@ -7,69 +7,44 @@ def Xform "ExampleModel" (
...
-                int[] faceVertexCounts = [4, 4, 4, 4, 4, 4]
-                int[] faceVertexIndices = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23]
-                point3f[] points = [(0, 0, 0), (1, 0, 0), (1, 1, 0), (0, 1, 0), (0, 0, 1), (1, 0, 1), (1, 1, 1), (0, 1, 1)]
-                # Additional high-detail attributes would be defined here
-            }
+            token visibility = "invisible"
+
+            # High detail geometry hierarchy goes here. It is invisible by
+            # default because it is not the default level of detail (see
+            # lod.default.index above).  Legacy renderers will not render it.
```
o2b3d, line 23: Typo. Attribute name is "lod:default:index" with colons rather than dots.

eleyaro:
I’m sure this was discussed at some point but I can’t see any info at first glance so sorry in advance for a duplicate question…  Is this proposal compatible with cluster-based LOD system similar to Nanite?

I imagine not since that kind of system may select many different LOD levels for a single mesh (depending on which parts are close to or far from the camera) and I assume that here level is selected per prim, so essentially per mesh.

meshula:
@eleyaro  If I were to approach meshlet based LOD systems such as nanite, I would approach it as follows:

Secondary meshlet data structures are necessary on the rendering side of the things to support nanite. Meshlets are not currently directly expressible, even though geometry subsets are immediately adjacent. 

Assuming you have some kind of post-process to associate USD source geometry with baked nanite geometry geometry: Take advantage of the floating point level of detail signal. The fractional component of the selection would key nanite into understanding the transition from near to far within the range.

spiffmon:
To address your statement: "that kind of system may select many different LOD levels for a single mesh", @eleyaro , there is nothing in our proposed schema that prevents a renderer from consuming multiple, or all, stated LOD levels.  A typical case for higher-quality transitions in some renderers would be to consume two adjacent levels at any one time and blend between them, based on a distance or screen-space metric.

eleyaro:
> there is nothing in our proposed schema that prevents a renderer from consuming multiple, or all, stated LOD levels

I understand, but let’s consider a case where the camera is placed on a flat, but very large and densely subdivided ground mesh. The camera can see both most detailed areas up close and very distant ones at the same time. Wouldn’t such case require loading all LODs for the ground mesh, effectively defeating the purpose of geometry streaming?

meshula:
The LOD schema selects amongst subgraphs, and the heuristic gives a specific renderer or engine an opportunity to provide a parameterized description of the inputs to its own heuristics that it can invoke as it renders. The value add is that you can save the parameters for your heuristic in a USD file, but in order to cover a system like nanite, view-adaptive tessellation, or streamed multi-resolution chunks (like Cesium's Tiles), it would make sense from my point of view that additional schemas might be necessary to carry extra data structures to describe those systems. 

Since this schema describes LODs as a selection, or combination, of sub graphs based on a heuristic that operates at a prim level, I'm not sure that there is directly a natural way, via subgraph selecting and combining, to specify your example of a densely subdivided ground mesh - the ground mesh ia in some sense a single entity, the same way that a nanite mesh is in some sense a single entity, even if it is made up of hundreds of meshlets.

Algorithms like ROAM and Nanite have a further constraint that complicates a simple heuristic selection.  In ROAM, in order to avoid cracks you have to introduce a secondary rule: If Node A selects a level of subdivision, its Base Neighbor B must also select a compatible subdivision.

Since nanite is its own DAG, the selection heuristic for DAG is actually an optimal graph cut through that DAG. It seems like that needs to be a specialized thing, rather than something that can be naturally expressed using existing USD graph structures.

To summarize my argument, there are two kinds of LOD, subgraph-switching LOD and spatial-structural LOD. One is about switching of subgraphs at discrete boundaries (and possibly combining them via e.g. a blending factor), the other is about optimized graph cuts of a single specialized entity, like a nanite mesh. USD's role is provide a standardized container for a data set and heuristic parameters for a renderer. My secondary argument would be that because something like nanite has specialized requirements, it would need to either be externally defined and evaluated in a renderer, or for interchange, a schema would need to be developed that explicitly describes it.

jesschimein:
Updated proposal to Implemented as the LoD schema is published and implemented as of 26.08! 
Note that LoD-related Hydra/imaging work is still in progress.

PR #82, opened 2025-05-05
Physics proposal for nested bodies

andrewkaufman:
### Description of Proposal

The intent of this proposal is to improve legibility of articulated rigid bodies in USD and to provide a more direct mapping when converting external datasets of physical articulations to USD.

We propose to amend the UsdPhysics Schema documentation & validation code to allow PhysicsRigidBodyAPI prims to have ancestor & descendant PhysicsRigidBodyAPI prims, even without the use of a resetXformStack XformOp.

This change for PhysicsRigidBodyAPI prims is conditional provided the nesting occurs within an articulation. The top-most ancestor PhysicsRigidBodyAPI must be, or have an ancestor which is, a PhysicsArticulationRootAPI.

[Link to Rendered Proposal](https://github.com/NVIDIA-Omniverse/USD-proposals/blob/physics_nested_bodies/proposals/physics_nested_bodies/README.md)

### Contributing

### DIFF
--- a/proposals/physics_nested_bodies/README.md
+++ b/proposals/physics_nested_bodies/README.md
@@ -4,8 +4,6 @@
 
 This proposal amends the UsdPhysics Schema documentation & validation code to allow PhysicsRigidBodyAPI prims to have ancestor & descendant PhysicsRigidBodyAPI prims, even without the use of a resetXformStack XformOp.
 
-This change for PhysicsRigidBodyAPI prims is conditional provided the nesting occurs within an articulation. The top-most ancestor PhysicsRigidBodyAPI must be, or have an ancestor which is, a PhysicsArticulationRootAPI.
-
 ## Background
 
 [UsdPhysics](https://openusd.org/release/api/usd_physics_page_front.html) is designed around rigid body simulators, which take as input a list of rigid bodies and a list of constraints. In particular, most design decisions are based on multi-body simulation, in which bodies are modeled as single, independent entities in world space, which by default move and respond to forces and torques independently. Constraints are used to reduce the degrees of freedom between two bodies.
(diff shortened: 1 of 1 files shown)


andrewkaufman:
I force pushed with a signed commit this time. No content changed.

andrewkaufman:
My understanding is that this prosal has been discussed in AOUSD Physics WG & has been accepted and the docs changes are being adopted into the WIP spec for UsdPhysics.

Given this, can someone with the necessary privledges update the status to greenlight the development work?

spiffmon:
@andrewkaufman , the intended flow for **new schemas** or substantive changes in behavior to existing ones is that it starts with a proposal (this), gets implemented in OpenUSD (likely via a PR, in this case), and **only then** is considered for inclusion in an AOUSD Spec to be published.  There are two reasons for this:

1. In the creation of AOUSD, we agreed it was important that OpenUSD be a reference, **superset** of OM's and behaviors that are normatively published, so that OpenUSD is reliable and definitive.
2. Part of OpenUSD's value-add (as opposed to something that officially exists only as a Spec) is that what's in it has some (or high, for most core schemas) stamp of "production battle tested".  It gives an opportunity for broad testing and some evolution, if needed, prior to being canonized in a Spec.

With that in mind, could I request that you incorporate the discussed changes into this proposal before we move this to "Published" ?  For now, I'm moving it to "Draft"

andrewkaufman:
@spiffmon that process outline makes total sense to me.

However, the issue seems to be somewhat out of my hands... I heard the AOUSD news independantly (I am not in the Physics WG currently) and only commented here to move the proposal process along & make sure we do get the implementation landed in the order you suggest.

> could I request that you incorporate the discussed changes into this proposal

The only feedback I have recieved is in casual conversation, and is to remove the "top-most ancestor PhysicsRigidBodyAPI must be, or have an ancestor which is, a PhysicsArticulationRootAPI" limitation. I have updated the proposal with that change now.

If there is other feedback you (or anyone) is aware of, please let me know.

PR #104, opened 2026-02-26
Adding Back Plate Support for USD

Denise-Yang:
### Description of Proposal

# Summary 

The majority of work in VFX is centered around augmenting existing footage, so it is important to view animation in the context of that footage (it doesn't matter if the animation looks good if it doesn't look integrated). We hope to more intuitive controls for VFX workflows by adding support for back plates and image planes to USD though we will leave the latter for when there is better infrastructure to implement image planes in Hydra.

[Link to Rendered Proposal](https://github.com/Denise-Yang/OpenUSD-proposals/blob/back-plates/proposals/back-plates/README.md)

### Contributing


jesschimein (approved):

jesschimein:
Updating proposal name to clarify that only Back plate support in USD has been added; Image Plane support is still to be worked on.

jesschimein:
Moved this proposal to Implemented as the Back Plates schema is ready to use as of 26.08. Back plates Hydra/imaging work is still in progress.

PR #105, opened 2026-03-08
Separation of Concerns for Identifiers in USD

asluk:
### Description of Proposal

# Summary

Standardize a mechanism for carrying external source identifiers on USD prims alongside namespace paths. Identifiers from systems like IFC, PLM, asset management databases, and telemetry platforms should round-trip through USD without loss or reliance on ad-hoc conventions.

# Problem statement

There is no standardized place in USD to store external identifiers. Pipelines work around this by encoding them into prim names, `customData`, or `displayName` -- all with significant trade-offs. The `displayName` migration into `uiHints` makes this time-sensitive.

# Glossary

- **Source identifier** -- An identifier from an external system (e.g., IFC GlobalId, PLM part number) independent of the USD namespace path.
- **Identity continuity** -- The requirement that source identifiers survive state transitions, re-parenting, and assembly changes.

# Reference Links

- [Extended Unicode Identifiers](https://github.com/NVIDIA-Omniverse/USD-proposals/tree/extended_unicode_identifiers/proposals/extended_unicode_identifiers) -- Complementary effort on prim name grammar
- [UI Hints in USD](https://github.com/PixarAnimationStudios/OpenUSD/tree/dev/pxr/usd/usdUI) -- `displayName` migration creating urgency
- [Composable Bindings](https://aka.ms/ComposableBindings) (Microsoft and NVIDIA, 2025) -- Digital engineering use case

# Details

Two distinct problems are often conflated: (1) an unencumbered source identifier field, and (2) improved prim name ergonomics. This proposal focuses on (1) without foreclosing (2). Two candidate approaches -- extending `assetInfo` with stratified sub-dictionaries vs. applied schemas with typed properties -- are compared with trade-offs and eight open questions.

# Risks

Uncurated proliferation, premature standardization, adoption fragmentation, and scope creep. See proposal for details.

# Alternate solutions

Existing mechanisms (`customData`, `displayName`, encoding into prim names) are evaluated in the proposal and found insufficient. The two candidate approaches are the alternatives under consideration.

# Out of Scope

Prim name grammar extensions, transcoding algorithms, feature-level identity continuity, and authorship traceability. Each is identified as complementary or downstream work in the proposal.

[Link to Rendered Proposal](https://github.com/asluk/OpenUSD-proposals/blob/aluk/source-identifiers/proposals/identifier_separation_of_concerns/README.md)

### Supporting Materials

None at this time.

### Contributing


proposals/identifier_separation_of_concerns/README.md
```diff
@@ -0,0 +1,904 @@
...
+   provides. What characteristics of the mechanism make that tractable?
+
+2. **Dictionary metadata vs. applied schema?**
+   Should source identifiers live as sub-dictionaries within `assetInfo`,
+   or as typed properties on applied schemas that contribute to
+   `UsdPrimDefinition`? (These are labeled Approach A and Approach B
+   respectively in [Likely direction](#likely-direction).) Key trade-offs:
+   - **Discoverability and validation**: applied schemas enable GUI
+     presentation of unauthored properties, schema versioning, and
+     schema-driven validation; dictionaries do not.
```
spiffmon: Wanted to note that if we are willing to stipulate that an organization/discipline/package that has identifier needs must provide a **fallback value** for each field they introduce (even if it be an empty string or assetPath), then an applied schema **can** introduce a subdictionary of `assetInfo` whose entries will be present in the UsdPrimDefinition, and therefore always discoverable (given GUI sensitive to the existence of `assetInfo`), just as properties would be.

If we adopt that path, we can also at least explore adding `usdGenSchema` support for generating getters/setters for metadata in the prim definition, though that might be a later goal.

asluk: Thanks @spiffmon  -- incorporated into https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/105/changes/ac9a3ac7a195f32391d4c4ec2345ec38edca38d1

proposals/identifier_separation_of_concerns/README.md
```diff
@@ -0,0 +1,519 @@
...
+**source name** from the originating system, which must be preserved exactly
+for round-trip fidelity.
+
+### Product Lifecycle Management (PLM) and Manufacturing
+
+PLM systems are the backbone of manufacturing workflows. They assign stable
+identifiers -- part numbers, revision codes, serial numbers -- that must
+persist across the lifecycle of a physical product and its digital twin.
+
+- A **part number** like `A-0000-12345-Rev.C` identifies the *design* of a
```
steveghee: Might i suggest we separate part number (A-0000-12345) from a specific revision. A part number by itself is often used as a generic reference ; searching for a part number may return a number of versions.  A specific revision might take a different form e.g. a Windchill part looks like this 'OR:wt.part.WTPart:4697800' where the details of the part are a. abstracted/hidden behind a more general purpose object model.  In this example, OR:wt.part.WTPart:4697800 might reference Rev.C of the part.  It might also represent a sub-assembly or some other item that isnt necessary given a part number.   Perhaps the easiest, most generic way to represent this is to describe the use of a more opaque 'handle' to a "object" in the PLM system.

proposals/identifier_separation_of_concerns/README.md
```diff
@@ -0,0 +1,519 @@
...
+  component. Multiple instances of that component in an assembly share this
+  identifier.
+- A **serial number** like `SN-2025-00847` identifies a specific *physical
+  instance* of that component.
+- A **BOM (Bill of Materials)** is generated by traversing an assembly and
```
steveghee, line 429: related to my comment above about object handles, another feature of PLM systems is that they can store and represent configurable products; a product structure (sometimes referred to as bill of materials) may be quite specific to the configuration of the product that is chosen.  To address a specific configuration, we might therefore have a second object reference or identified which we call the "navigation criteria" - this is effectively the set of rules to use to traverse and return the (sub) structure of a product - it's more complex than a pure 'filter' as navigation of a structure can actually add or change (even move) objects.  The BOM can therefore change quite dramatically based on the criteria chose.   Basically what i'm getting at there is that there may not be a single 'key' that can identify the specific item that we are referencing in the USD ; you mention this in the preamble above, so i thought this could be a good example of the use of / need for multiple keys.  In a general sense, the navigation criteria (for windchill) can be an object in its own right e.g OR:wt.filter.NavigationCriteria:7608531, or it could be a string (encoded URL or JSON).  Perhaps safe to say its generically a string/token.  Other systems may have other ways of doing this, and have additional keys.

proposals/identifier_separation_of_concerns/README.md
```diff
@@ -0,0 +1,519 @@
...
+   it be extended with standardized keys for source identifiers from multiple
+   systems? Or would overloading `assetInfo` conflate USD asset identity with
+   external system identity?
+
+2. **Schema or metadata?**
+   Should source identifiers be expressed as an applied API schema (like
+   `SemanticsAPI`), as metadata (like `assetInfo`), or as properties? Each has
+   different implications for composition, queryability, and performance.
+
+3. **Scope: model roots only, or any prim?**
```
steveghee: my 2c here.  *any prim* would be my preference; an example usecase where  a generic "xref" ID can be valuable is when you have (for example) a SimReadified asset where certain choices have been made to simplify/collapse the representation e.g. hide the internals, so the visual representation (prims) may "look" like the item but the structure has been collapsed to remove the actual product structure. There is still value, however, in retaining IDs of the remaining visual assets  e.g. this prim/mesh/subgeom is a subset of this <part>.  This is important for composite products e.g. an assembly (rack) containing a collection of components (trays).

proposals/identifier_separation_of_concerns/README.md
```diff
@@ -0,0 +1,519 @@
...
+   `assetInfo` is scoped to model roots. External identifiers are needed on
+   prims at all levels of the hierarchy -- individual rooms, structural
+   members, electrical components, fasteners. The mechanism must not be
+   artificially limited to model roots.
+
+4. **Namespacing of identifiers.**
```
steveghee: i think you'd want a bit of both. In general, i would argue there's a "primary" identifier key that virtaually everythin would need. the question is whether you use namespacing to separate this out to different domans e.g. would a windchill primary be ptc:windchill:identfier=xyz, or should the vendor/domain specifics be encodable IN the identifier e.g. identifer = "com.ptc.windchill:part:xyz".  I prefer the latter.  Outside of the primary key, anyone who needs a secondary key is most likely going to want to name it because its more than likely going to by fairly specific to them; this is where the namespacing comes in e.g. you could have primary=xyz, ptc:windchill:secondary=abc.  The only reason i can think for namespacing on the primary is if an asset can potentially have multiple primary IDs e.g. maybe i can find it (or pieces of it) in multuple systems. I guess the arguement there might be again to stick to a single primary, but allow this to be an array e.g. identifier=["windchill:x","sap:y","salesforce:z"]. I guess the array also allows for non-destrictive composition e.g. someone could add a reference that they may need, but it doesnt remove the original.

proposals/identifier_separation_of_concerns/README.md
```diff
@@ -0,0 +1,519 @@
...
+   If a prim carries identifiers from multiple external systems, how should
+   they be organized? A flat dictionary with well-known keys? A
+   multiply-instanced schema with system-specific instance names?
+
+5. **Relationship to `displayName`.**
+   How does the source identifier relate to the prim's `displayName`? In some
```
steveghee: my 2c, there is no explicit relationship to displayName, which may well be derived from information that is stored in the referenced item, but the displayName provided should have no influence on the identifer. The identifier is the source of truth.  displayName is a UX convenience.

proposals/identifier_separation_of_concerns/README.md
steveghee: some other examples for multiple identifiers are alternative parts e.g. OEM partX and replacement partY - typically not the same part, but form/fit/function equivalent.  A similar example is service part numbers and kits, where again there are different ID schema used for the same base part - sometimes the service part number allows for OEM/replacement parts. Sometimes, spare parts may be available directly, or within a special kit (colllection of parts to be replaced).  I'm providing these are more examples of usecases that highlight why an identifier array might be valuable.

proposals/identifier_separation_of_concerns/README.md
asluk: Thanks @steveghee  -- I took a pass in https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/105/changes/2cc91d8731ef2a2093cbf7d9626ef612412e8e7f

proposals/identifier_separation_of_concerns/README.md
asluk: Thanks @steveghee  -- I took a pass in https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/105/changes/2cc91d8731ef2a2093cbf7d9626ef612412e8e7f

proposals/identifier_separation_of_concerns/README.md
asluk: Thanks @steveghee  -- I took a pass in https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/105/changes/2cc91d8731ef2a2093cbf7d9626ef612412e8e7f

proposals/identifier_separation_of_concerns/README.md
asluk: Thanks @steveghee  -- I took a pass in https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/105/changes/2cc91d8731ef2a2093cbf7d9626ef612412e8e7f

proposals/identifier_separation_of_concerns/README.md
asluk: Thanks @steveghee  -- I took a pass in https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/105/changes/2cc91d8731ef2a2093cbf7d9626ef612412e8e7f

proposals/UsdSolid/UsdSolidBrep.md
```diff
@@ -0,0 +1,1471 @@
...
+This preliminary schema has been reviewed by AOUSD Geom WG and will continue to be iterated upon in a public draft proposal.
+We are publishing the schema in this preliminary form to share our current findings and take the next step in addressing uses of the design.
+By making the schema public we hope to engage the broader CAD/CAE community, kicking off efforts to produce datasets using this schema to further unlock exploratory work for downstream integration.
+Potential use cases are recorded [here](https://docs.google.com/document/d/1kXiZPhkUh6JfohE2Q7qVA4hSuJD3UyiBdYW_nEsEIJU/edit?tab=t.0)
+
+
+Adding Breps to OpenUSD creates questions with answers outside the scope of the Geometry WG. 
+For example, we hope to address the ways in which data can be connected in USD.
+Breps and Meshes are a source and derived data pair that don't have a natural connectivity in OpenUSD.
+A Brep prim should know if it has a Mesh on the stage, so it doesn't have to be tessellated.
```
steveghee: here's a thought on this ; its generally the case that there is a N:1 mapping of gprims to brep. because the gprims will (should be) derived from the brep source, a mechanism for implementing this could be to back reference the brep id (by id or path) from the gprim(s).  Perhaps this id xref (id array) can be added to gprims via an API schema.   Of course at runtime, you load and xref the gprims back to the brep, and now the brep knows which gprims it is responsible for, so you can resolved line 18 at runtime.  Line 19 is resolved by the gprim (e.g. mesh) referencing the brep. Line 20 is resolved by gprims being able to reference multiple breps.  Thoughts?

asluk: oops this accidentally got put in the wrong PR , @steveghee  -- I've moved your comment to https://github.com/aousd/Geom-WG-OpenUSD-Proposals/pull/8/changes#r2996427305 - please feel free to correct and continue there.

proposals/identifier_separation_of_concerns/README.md
```diff
@@ -0,0 +1,1030 @@
...
+
+#### Configurable products and composite keys
+
+PLM systems also manage **configurable products** whose structure depends
+on context. A product's Bill of Materials may change dramatically based on
+**navigation criteria** -- the rules used to traverse and resolve the
+sub-structure for a specific configuration. Navigation criteria may
+themselves be opaque objects (e.g., Windchill's
+`OR:wt.filter.NavigationCriteria:7608531`) or encoded strings (URLs,
+JSON). The result is that a single primary identifier may not suffice to
```
perfectproducts, line 466: Another example would be the mercedes benz part system with the actual part number "A-PartNumber"(https://www.benzworld.org/threads/the-mercedes-benz-parts-numbering-system.1473937/) and additional extension codes "ES1", "ES2" for color information, also the full number follows a rule set (use base number only or with either ES1 or ES2 suffix number, color code in ES2 is four digits.  This hints to me to using a schema because it would be nice to be able to verify if its a correct identifier in the source system

asluk: Thanks! Incorporated in https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/105/changes/c33200ae6902843d62a9d2595f9e9c2f1aebc157

PR #109, opened 2026-05-20
Solid Models in USD

asluk:
### Description of Proposal

# Summary

Introduce **UsdSolid**, a boundary representation (Brep) schema for OpenUSD built on the Radial Edge Data Model. The schema supports manifold and non-manifold solids, sheet bodies, and wireframes, with explicit topological relationships and a packed `UsdSolidBrepArray` design that limits prim count for large assemblies.

This is the technical companion to the *Boundary Representation Geometry in OpenUSD — Problem Statement* proposal (PR #108); please review that proposal first for use cases and motivation. Both proposals are opened against `main`, so this PR's diff includes the P1 problem-statement file as well as the P2 schema files until P1 merges; reviewers may wish to focus on `proposals/UsdSolid/` here.

This proposal is from the AOUSD Geometry Working Group.

# Problem statement

OpenUSD has no native representation for mathematically exact solid geometry. The companion problem statement develops the use cases and design considerations in depth. In brief, manufacturing, AECO, robotics, simulation, and digital-twin workflows require exact Brep geometry that mesh and subdivision-surface types cannot express. This proposal provides a candidate schema.

# Glossary

The full Brep glossary lives in the companion problem statement's Appendix A; this proposal's Glossary section adds only the schema-specific terms:

- **BrepArray** -- A `UsdSolidBrepArray` prim that holds one or more Breps in a packed, flat representation. Each Brep's topology, geometry, and metadata are concatenated into shared arrays and delineated by per-Brep counts and offsets.
- **Wire edge** -- A `wireEdge` is a topologically standalone edge that is not part of any face loop. Wire edges carry the same curve geometry as loop-bounding edges but exist independently as one-dimensional features within a shell.
- **Radial Edge Data Model** -- A non-manifold-capable topological data model (Weiler, 1986) that supports manifold solids, non-manifold solids, sheet bodies, and wireframes through a unified "use" abstraction (edgeuse, faceuse).

# Reference Links

- **Companion problem-statement proposal** -- PR #108
- [Weiler, K. (1986). *The Radial Edge Structure: a Topological Representation for Non-Manifold Geometric Boundary Modeling*](https://webserver2.tecgraf.puc-rio.br/~lfm/teses/KevinWeiler-Doutorado-1986.pdf) -- The topological model the schema is built on.
- [PRC specification (ISO 14739-1:2014)](https://www.iso.org/standard/54845.html) -- Reference for the geometry-type catalog the schema targets.
- [Gaussian Splats schema (OpenUSD PR #3716)](https://github.com/PixarAnimationStudios/OpenUSD/pull/3716) -- A precedent for a preliminary geometry schema (`prelimUsdRender...`) entering OpenUSD and later promoting.
- [USD Profiles proposal (PR #89)](https://github.com/PixarAnimationStudios/OpenUSD-proposals/tree/main/proposals/profiles) -- A relevant mechanism for declaring which capabilities a USD implementation supports.

# Details

The proposal is organized as **Introduction → Glossary → Design → Schema → Examples → References**.

**Design** introduces the Brep data model in three layers — shape, topology and "use", and Brep — and then walks through USD-implementation choices including `UsdSolidBrepArray`, instancing, geometry typing, modeling Breps on a UsdStage, and trimming curves. Subsections on *Flexible design possibilities* and *Other implementations considered* document the design space the proposal explored before converging on the Radial Edge model and a packed BrepAPI on a typed prim. Assemblies, tolerance, validation, risks, and open questions follow.

**Schema** is a complete `prelimUsdSolid` schema definition (`prelimUsdSolid` prefix and `skipCodeGeneration = true`, following the OpenUSD convention for WIP schemas; both will be stripped on acceptance, matching the path Gaussian splats took).

**Examples** narrate five `.usda` cases covering cube, non-manifold cubes, cube with internal void, BrepArray with multiple Breps, and per-face materials.

Design choices worth highlighting for reviewers:

- **Radial Edge over winged-edge / half-edge** -- chosen for explicit non-manifold support, which industrial assemblies require. The topology models in PRC, STEP, and Parasolid map into the Radial Edge model.
- **Packed BrepArray** -- a `UsdSolidBrepArray` typed prim with a multiple-apply `UsdSolidBrepAPI` keeps prim counts tractable for large assemblies. The trade-offs against one-prim-per-Brep and one-prim-per-geometry-object alternatives are documented in the Design section.
- **Double-precision geometry** -- proposed for NURBS control vertices, weights, and knots (and analytic surface parameters), addressing a known limitation of existing `UsdGeomNurbsCurves` / `UsdGeomNurbsPatch` for CAD interchange.
- **PRC geometry-type catalog as target** -- the proposal lists the full set of analytic curves, surfaces, and volumes (Helix, Revolution, Extrusion, Torus, etc.) that the schema is structured to accommodate. NURBS land first; the analytic types are an extension path, not a v1 deliverable.

# Risks

From the proposal's Risks subsection:

- **Validation complexity.** Full geometric validation (self-intersection, tangency, tolerance enforcement) requires a geometry kernel and is outside the scope of OpenUSD. Topological validation is in scope; geometric validity depends on third-party libraries.
- **Geometry kernel dependencies for rendering.** Tessellation of Breps for rendering requires a kernel. OpenUSD does not ship one; adoption depends on the ecosystem (open source or commercial kernels) consuming UsdSolid data.
- **Performance at scale.** Large assemblies (thousands of Breps, tens of thousands of faces) drove the packed `UsdSolidBrepArray` design, but composition, layering, and instancing performance remain to be characterized with production datasets.
- **Ecosystem readiness.** DCCs, viewers, and engines will need to add Brep support before the schema delivers end-user value.
- **Data quality dependence on exporters.** Authoring happens outside OpenUSD; quality tracks exporter quality. The WG is considering reference datasets and a conformance suite.

# Alternate solutions

The proposal's *Other implementations considered* subsection documents the alternatives evaluated:

- **One UsdPrim per geometry object** -- rejected on prim-count grounds at industrial-assembly scale.
- **One UsdPrim per Brep** -- rejected for the same reason; assemblies routinely contain thousands of Breps.
- **Breps as an Applied API on existing prim types** -- considered; the typed-prim approach with a multiple-apply API was preferred for stronger schema semantics on the container.
- **Breps as a black box (opaque file reference)** -- rejected because it forfeits USD's composition, layering, and per-element property assignment. The companion problem statement develops the full rebuttal.

# Out of Scope

- **Detailed schemas for analytic geometry types** -- the catalog is identified (Helix, Revolution, Extrusion, Torus, etc.) and the schema is structured to accommodate them, but their detailed designs are deferred to a follow-up revision.
- **Brep <-> Mesh correlation** -- identified as an open question. A candidate approach (gprim back-references via an applied API) is sketched in Open Questions; full design is deferred until the core schema is in use.
- **CAD assemblies as a schema** -- whether assemblies warrant a dedicated schema or a best-practices guide built on instancing and constraints is unresolved; covered in Open Questions.
- **Constraint systems for assembly simulation** -- mates, joints, kinematic constraints; identified as open. May belong here, in a separate schema, or in an OpenUSD-wide proposal.
- **Migration tooling from STEP, PRC, Parasolid** -- detailed mappings and exporter implementations are downstream work.
- **Implementation in OpenUSD** -- staged separately, modeled on the Gaussian splats path (`prelimUsdRenderGaussian` → `UsdRenderGaussian` after community use).

[Link to Rendered Proposal](https://github.com/aousd/OpenUSD-proposals/blob/p2-usdsolid-schema/proposals/UsdSolid/README.md)

### Supporting Materials

The proposal includes six diagrams (`proposals/UsdSolid/images/`) illustrating the Radial Edge data model, BrepArray structure, and face-material binding. Five narrated `.usda` examples accompany the schema.

### Contributing


dgovil:
Hi Aaron,
I am just relaying some internal feedback:

Can we reconsider deferring the correlation between the BRep and the tessellated mesh? These are the two main concerns:

**Vendor divergence risk:** If we ship without a standardized correlation mechanism, individual CAD vendors might develop their own conventions to bridge the two representations. Once these conventions become widely adopted and assets rely on them, standardizing later becomes a migration issue rather than a design challenge. The cost of getting this right upfront is significantly lower than the cost of reconciling multiple vendor-specific schemes later.

**Tessellated mesh may not always be present:** CAD vendors might choose not to ship a tessellated mesh at all if the correlation path is unclear. If the contract for going BRep - mesh isn’t clearly defined, the safest choice for a new vendor is to ship BRep only and let consumers tessellate. This is arguably a worse outcome than having a dual representation, and it’s the default we’ll end up with if we defer.

asluk:
> Hi Aaron, I am just relaying some internal feedback:
> 
> Can we reconsider deferring the correlation between the BRep and the tessellated mesh? These are the two main concerns:
> 
> **Vendor divergence risk:** If we ship without a standardized correlation mechanism, individual CAD vendors might develop their own conventions to bridge the two representations. Once these conventions become widely adopted and assets rely on them, standardizing later becomes a migration issue rather than a design challenge. The cost of getting this right upfront is significantly lower than the cost of reconciling multiple vendor-specific schemes later.
> 
> **Tessellated mesh may not always be present:** CAD vendors might choose not to ship a tessellated mesh at all if the correlation path is unclear. If the contract for going BRep - mesh isn't clearly defined, the safest choice for a new vendor is to ship BRep only and let consumers tessellate. This is arguably a worse outcome than having a dual representation, and it's the default we'll end up with if we defer.

Hi @dgovil — thanks for surfacing this. Responding for whoever raised it:

Correlation isn't deferred — it's a separate concern from the Brep schema. The companion problem statement's [Coexistence with meshes principle](https://github.com/aousd/OpenUSD-proposals/blob/p1-exact-geometry-problem-statement/proposals/cad_geometry/README.md#principles) already says Brep and Mesh "should be relatable… but neither should require the other," and the mechanism is "separable from the core Brep schema." This proposal's [Open Questions](https://github.com/aousd/OpenUSD-proposals/blob/p2-usdsolid-schema/proposals/UsdSolid/README.md#open-questions) section has a "Brep↔Mesh correlation" entry sketching a candidate from Steve Ghee: an applied schema on the *derived* gprim that back-references the source Brep — the active follow-on path.

Two reasons it sits there:

- **Broader than Brep↔Mesh.** Source/derived pairs aren't unique to CAD — subdivision cages and limit surfaces, sim results and viz meshes, LoD chains. A mechanism designed only for UsdSolid would need rework when the next pair appears. It belongs in its own proposal, seeded by prior discussions.
- **USD composition supports the separation.** The correlation mechanism sits outside UsdSolid's domain entirely; the core Brep schema stays self-contained.

On the two specific concerns:

**Vendor divergence.** The Ghee sketch signals direction. Vendor-prefixed proposals exploring variations are welcome — AOUSD's vendor → multi-vendor → core pattern (cf. [PR #105](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/105)) converges rather than fragments. Symmetric risk: continuing to block the Brep schema leaves vendor-specific Brep encodings hardening in `customData` and opaque payloads — a divergence we're already paying for.

**Brep without companion mesh.** The problem statement frames this as a desired direction, not a worse outcome — see ["'Derive on demand' is replacing pre-tessellation"](https://github.com/aousd/OpenUSD-proposals/blob/p1-exact-geometry-problem-statement/proposals/cad_geometry/README.md#why-this-matters-now) and [Hero Use Case 2](https://github.com/aousd/OpenUSD-proposals/blob/p1-exact-geometry-problem-statement/proposals/cad_geometry/README.md#hero-use-case-2-consumer-controlled-data-quality) (consumer-controlled data quality). The schema doesn't preclude shipping both; the question is whether the correlation *contract* must land in v1, and we don't think so. Real data on real stages is how the community converges on the right mechanism.

Best,
@asluk and @umhoefer

martinpwnv:
Could I suggest moving the schema.usda out of the main doc where it is a bit hard to find (currently in a collapsible section) and have it be a standalone file in the repo that lives alongside the document?

asluk:
Good call, Martin — done. The schema now lives as a standalone [`schema.usda`](https://github.com/aousd/OpenUSD-proposals/blob/p2-usdsolid-schema/proposals/UsdSolid/schema.usda) alongside the document, and the README links to it instead of burying it in a collapsible section. The content is byte-identical to the previously inlined block (no edits beyond the move). Pushed as 811f03c.

asluk:
**Re: `schema.usda` extraction — tradeoff + a future idea**

Moving the schema into a sibling [`schema.usda`](./schema.usda) is the approach we've landed on for now. It has real tradeoffs worth noting for the record:

**Wins**
- Single canonical, machine-readable copy (good for tooling/codegen)
- No drift between an inlined block and the real schema
- Easy to find, lives alongside the doc

**Cost**
- Schema no longer visible *in place* in the proposal
- Relative `./schema.usda` link works on GitHub's web view, but **breaks in portable/offline renderings** — a PDF or static export (e.g. Pandoc) can't follow it, so "read the proposal on the go" hits a dead reference

**Future work for consideration**
- If "canonical file + reference from prose" becomes a common pattern, a small **proposal build step** could transclude the file into the rendered doc (marker comments + CI staleness check) — one source of truth *and* self-contained artifacts (GitHub / PDF / offline)

Flagging only; not proposing we act on it here.

proposals/UsdSolid/schema.usda
```diff
@@ -0,0 +1,391 @@
...
+    )
+} # end class "BrepPointAPI"
+
+#************************************************************************************
+# purpose: curve geometry apis for BrepArrays 
+# note   :  BrepCurveUvNurbAPI: singleApply API = associated geometry of edgeuse objects defined by 2d BSpline curves
+#           
+#           BrepCurve3dNurbAPI  : multipleApply API = associated geometry of edges defined by 3d BSpline curves
+#             Instance name = edgeCurve3dNurb   => associated 3d NURB curve geometry of edge objects
+#             Instance name = wireEdgeCurveNurb => associated 3d NURB curve geometry of wireEdge objects
```
donmccurdy, line 160: Possibly typos or out of date? Mismatched with schema below:


```suggestion
#             Instance name = edge3dNurb   => associated 3d NURB curve geometry of edge objects
#             Instance name = wireEdge3dNurb => associated 3d NURB curve geometry of wireEdge objects
```

proposals/UsdSolid/schema.usda
```diff
@@ -0,0 +1,391 @@
...
+        }
+    )
+    
+    uniform double[] curve3d:nurb:weights (
+        doc = """ Curve3d_ii's "w" weight components for each control vertex. 
+        Must be the same length as the curveControlVertices attribute. Weights must be positive, w>0. 
+        \\note Some DCC's pre-weight the \\em points, but in this schema, \\em points are not pre-weighted.        
+        size() = curve3dControlVertices.size() """
+        customData = {
+            string apiName = "Curve3dWeights"
```
donmccurdy, line 229: Nit: case inconsistency with APIs above.


```suggestion
            string apiName = "curve3dWeights"
```

proposals/UsdSolid/README.md
```diff
@@ -0,0 +1,931 @@
...
+#### Instancing of Brep models
+
+In this proposal, whole _UsdSolidBrepArray_ can be referenced to create multiple instances of a set of Breps.
+
+#### Brep geometry in USD
+
+There are 4 types of geometry stored along with the _UsdSolidBrepArray_.
+The simplest is the vertex location, which is stored as a point3d.
+An edge needs a curve and a face needs a surface to have shape, so curves and surfaces are applied APIs, where owning Edges and Faces indicate which geometry gives them shape.
+UVTrimCurves are the fourth geometry object, also an applied API.
```
donmccurdy: Apologies if I've missed it, here or elsewhere in OpenUSD references, but—I'm having trouble finding a definition for the term "UVTrimCurve" as used in this document?

proposals/UsdSolid/README.md
```diff
@@ -0,0 +1,931 @@
...
+#### Modeling Breps on a UsdStage
+
+Efficient modeling or editing a Brep directly on a UsdStage is not a feature of the current schema, but there are clear steps to take to support this.
+Creating schemas for the 11 topology and use objects in the Brep model will allow the entire directed graph structure of the Brep to be represented in USD, which will be well-suited for live editing.
+
+#### Trimming curves
+
+Optional trim curves can be included similarly to curve and surface geometry.
+The edgeuse defines the connection between a given edge and face; the edgeuse stores an index for the associated trimming curve applied API.
+Trim curves are optional in this Brep model because the edge curve defines the model truth, but trim curves are useful for, e.g., speeding up tessellation algorithms.
```
donmccurdy, line 271: Apologies if I've missed something—I'm having trouble following the definition for "UVTrimCurve", and how an edgeuse is associated to its trim curve. Is that fully defined in this document and `schema.usda`?

I assume the intention is that `brep:curveUv:nurb:*` holds 0–1 trim curves per edgeuse, but how might it be represented if some edgeuse instances have trim curves and others do not?

proposals/UsdSolid/schema.usda
```diff
@@ -0,0 +1,391 @@
...
+        The length of the slice of the array for the iith curve must be ( vertexCount[ii] + order[ii] ), and its entries
+        must take on non-decreasing values. Knots are listed in multiplicity, e.g. [0, 0, 1, 1]."""
+        customData = {
+            string apiName = "curveUvKnots"
+        }
+    )
+    
+    uniform double[] brep:curveUv:nurb:weights (
+        doc = """ CurveUv_ii's "w" weight components for each control vertex. 
+        Must be the same length as the edgeCurveControlVertices attribute. Weights must be positive, w>0.
```
donmccurdy, line 280: Probably a reference to `curveUvControlVertices`?

```suggestion
        Must be the same length as the curveUvControlVertices attribute. Weights must be positive, w>0. 
```

proposals/UsdSolid/README.md
```diff
@@ -0,0 +1,931 @@
...
+        uniform token[] edgeuse:orientationType = ["same", "same", "opposite", "opposite", "opposite", "same", "same", "opposite", "same", "same", "opposite", "opposite", "opposite", "same", "same", "opposite", "opposite", "same", "same", "opposite", "same", "same", "opposite", "opposite"]
+        uniform token[] edgeuse:thisRadialEntryType = ["topEntry", "topEntry", "bottomEntry", "bottomEntry", "topEntry", "bottomEntry", "topEntry", "topEntry", "bottomEntry", "bottomEntry", "topEntry", "bottomEntry", "bottomEntry", "topEntry", "topEntry", "bottomEntry", "topEntry", "topEntry", "topEntry", "topEntry", "bottomEntry", "bottomEntry", "bottomEntry", "bottomEntry"]
+        uniform float3[] extent = [(0, 0, 0), (1, 1, 1)]
+        uniform uint[] face:loopCount = [1, 1, 1, 1, 1, 1]
+        uniform double2[] face:range = [(0, 0), (1, 1), (0, 0), (1, 1), (0, 0), (1, 1), (0, 0), (1, 1), (0, 0), (1, 1), (0, 0), (1, 1)]
+        uniform token[] face:surfaceType = ["BrepSurfaceNurbAPI", "BrepSurfaceNurbAPI", "BrepSurfaceNurbAPI", "BrepSurfaceNurbAPI", "BrepSurfaceNurbAPI", "BrepSurfaceNurbAPI"]
```
donmccurdy, line 533: Question - here, where all N entries in `face:surfaceType` are of the same type, the face `i` maps exactly to surface `i` in `brep:surface:nurb:*`. If the BrepArray contained faces of mixed surface types (e.g. in a future version with analytic surfaces), how is the face/surface association resolved? Does each face effectively claim the next 'unclaimed' surface in the arrays of the corresponding surface type?

donmccurdy (commented):
A few questions and comments, inline!

proposals/UsdSolid/README.md
umhoefer: Good catch.  We need to update the docs.  The intent is that for "no curve" at index `i` you can set `brep:curveUv:nurb:vertexCount[i]=0` and `brep:curveUv:nurb:order[i]=0`.  The other attributes can be left with no values for that curve.

proposals/UsdSolid/schema.usda
umhoefer: Thanks. We'll fix the documentation typos you pointed out.

proposals/UsdSolid/README.md
umhoefer: Exactly.  So if we had a Brep with `face:surfaceType = ["NURB", "Cylinder", "Cone", "NURB"]` then the correspondence is 
```
face_0 <-> NURB_0
face_1 <-> Cylinder_0
face_2 <-> Cone_0
face_3 <-> NURB_1
```
I'll add this to the docs.

proposals/UsdSolid/README.md
donmccurdy: Thanks @umhoefer! With the explanation I think that makes sense.

PR #111, opened 2026-06-23
Deformable Body Physics in USD

sschirm:
#### Summary

This proposal extends `UsdPhysics` with a basic, interoperable, backwards-compatible set of abstractions for **deformable body dynamics** — the elastic deformation of volumes, surfaces (e.g. cloth, shells), and curves (e.g. hair, rods). It mirrors the design philosophy of the 2021 rigid body schema (real-world material quantities, unit independence, properties on materials, adding APIs to existing classes over introducing new prims) and interoperates with rigid bodies so both can coexist in a stage.

#### Problem statement

USD has expressed rigid body physics since 2021, but there is no standard way to represent deformable content — the materials, simulation geometry, rest shape, collision, embedding, and attachment data needed to simulate it. This proposal defines those abstractions so deformable content can be exchanged across DCCs and simulators, the way rigid bodies are today.

#### Scope

In scope (new `UsdPhysics` types — see the proposal for the full list and details):

- A shared `UsdPhysicsBodyAPI` (rigid body refactor, backwards compatible).
- Deformable materials and simulation APIs for volume, surface, and curve deformables.
- Deformable bodies, simulation geometry, and rest shape.
- Collision and graphics geometries, and bind poses for geometry embedding.
- Material assignment and mass distribution.
- Kinematic deformables, attachments, and element collision filtering.

Out of scope (left for future extensions):

- Plastic deformation.
- Per-point / per-element kinematic state.
- Direct rotational coupling in attachments.
- Continuous (primvar-driven) material variation.

#### Reference links

- Rendered proposal: [wp_deformable_physics.md](https://github.com/aousd/OpenUSD-proposals/blob/physics-deformables/proposals/physics_deformables/wp_deformable_physics.md)
- Folder overview: [README.md](https://github.com/aousd/OpenUSD-proposals/blob/physics-deformables/proposals/physics_deformables/README.md)
- Developed in the AOUSD Physics Working Group ([aousd.org](https://aousd.org)); builds on the existing `UsdPhysics` rigid body schema (2021).

_Design rationale, risks, and alternatives considered are documented inline in the proposal's Design Notes._

### Supporting Materials

The proposal includes `.usda` examples throughout and embeds illustrative figures under [`proposals/physics_deformables/images/`](https://github.com/aousd/OpenUSD-proposals/tree/physics-deformables/proposals/physics_deformables/images).

### Contributing

### DIFF
--- a/proposals/physics_deformables/wp_deformable_physics.md
+++ b/proposals/physics_deformables/wp_deformable_physics.md
@@ -1,0 +1,904 @@
+# Deformable Body Physics in USD Proposal
+
+- [Purpose and Scope](#purpose-and-scope)
+- [Guiding Principles](#guiding-principles)
+- [USD Implementation](#usd-implementation)
+  - [Scenes](#scenes)
+  - [Rigid Body Refactor](#rigid-body-refactor)
+  - [Physics Material](#physics-material)
+  - [Deformable Materials](#deformable-materials)
+  - [Deformable Bodies](#deformable-bodies)
+  - [Simulation Geometry and Rest Shape](#simulation-geometry-and-rest-shape)
+  - [Rest Shape Attributes](#rest-shape-attributes)
+  - [Collision Geometries](#collision-geometries)
+  - [Graphics Geometries](#graphics-geometries)
+  - [Geometry Embeddings](#geometry-embeddings)
+  - [Assigning Materials](#assigning-materials)
+  - [Mass Distribution](#mass-distribution)
+  - [Kinematic Deformables](#kinematic-deformables)
+  - [Attachments](#attachments)
+  - [Element Collision Filtering](#element-collision-filtering)
+
+## Purpose and Scope
+
+After extending USD with the ability to express rigid body physics simulation concepts in 2021, the USD Physics working group has decided to prioritize further extending USD with concepts to express deformable body dynamics. This specification defines a similarly basic set of abstractions for expressing the initial conditions for the simulation of elastic deformation, including that of volumes, surfaces (e.g. cloth, shells) and curves (e.g. hair, strands). Concerns around plastic deformation were left for a future extension.
+
+Fewer generally available and commonly used tools and libraries exist for deformable simulation than for rigid body simulation, which made the search for a common set of abstractions more difficult. The same assumptions about target applications that were made for the USD Physics rigid bodies schema are made for deformables too: that the schema is primarily meant to serve interactive 3D graphics applications, and that users will be adding simulation capabilities to existing graphics-only USD content. A seamless extension of USD is intended, with the new simulation capabilities being made to interoperate with the existing rigid body abstractions, so that soft objects, cloth, hair, or fur can be represented and believably simulated.
+
+> **Design Note**
+>
+> Initially, it was planned to modify the original USD Physics schema for rigid bodies to integrate deformables in a way that included significant refactors of the existing schema to combine commonalities. Ultimately, the importance of formulating the USD Physics deformable bodies schema using minimal changes in a manner that preserves backwards compatibility was recognized. This approach ensures that both current and future applications can load existing assets as well as those that include deformables.
+
+## Guiding Principles
+
+As for the USD Physics rigid bodies schema, the intention was to focus on real-world concepts rather than on simulation-specific abstractions, unless those abstractions were overwhelmingly common and helpful. In particular, real-world engineering quantities are used for material properties. USD's preference for unit independence is maintained. The pattern established in the USD Physics rigid bodies schema is also continued, in which material properties are added to materials rather than to the simulated objects directly. The concept of physics scenes is retained. The USD design principle of preferring the use of existing classes or the addition of APIs to existing classes before introducing new prims is also adhered to.
+
+As with rigid bodies, all physics objects, now including deformables, automatically collide and interact by default.
+
+## USD Implementation
+
+### Scenes
+
+The USD Physics rigid bodies schema established the concept of physics scenes to enable multiple independent simulations within a single stage with different settings for fundamental properties such as gravity. Like rigid bodies, deformables conceptually simulate in a scene. Everything in a scene interacts by default, and things in different scenes do not interact. The gravity of a scene is applied to all the deformables in the scene. Deformables are associated with a scene in the same manner as rigid bodies, using the `simulationOwner` attribute.
+
+### Rigid Body Refactor
+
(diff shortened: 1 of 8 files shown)


sschirm:
Two Suggested Changes Regarding Material Stiffnesses
---------------------------------------------------------------------------------

[stiffness_modulus_reference.pdf](https://github.com/user-attachments/files/29465385/stiffness_modulus_reference.pdf)

**Stretch, shear, bend and twist stiffness units and nomenclature**
 
Newton Cable Sim developer pointed out that the proposal either has stretch, shear, bend and twist stiffness attributes units wrong (force/area) or their naming is misleading. There are two options to address this:

1. Use correct stiffness units according to energy density equation (see attached PDF)
In this variant, the geometric thickness is ignored (thickness would then only matter to translate volumetric density to masses).
The stiffnesses can likely be consumed directly by the simulator. They are probably more familiar to developers and content creators.

2. Rename to stretchModulus, shearModulus, bendModulus, twistModulus and keep current force/area units.
In this variant, effective stiffness values get scaled by the geometric thickness according to specified rules (see attached PDF), leveraging the thickness attribute further. It might be nice to tweak thickness and having stiffnesses reasonably scaled. As opposed to variant A) , the assumption of a circular cross-section for curve deformables has to be maintained (which is already the case for mass distribution).

While 1) is more direct, more familiar, and avoids the circular cross-section assumption, 2) is more aligned with the current reliance on the geometric thickness specification. For completeness I would like to mention that another Newton Deformables developer, also provided the feedback that the value of the thickness scaling mechanism is questionable. The developer would have preferred to have separate density definitions with corresponding units for surfaces and curves, making "thickness" redundant. This suggestion seems to go against inheriting the PhysicsMaterial density, which should maintain its volumetric interpretation.

**Young's Modulus and Poisson's Ratio fallback attributes**

LightWheel developers provided another but related feedback. They like the ability to specify per deforming-mode stiffnesses through isotropic material properties (Young's and Poisson's) also for surfaces and curves, as we used to propose as well. We steered away from this IIRC for two reasons:

1. We clarified and simplified the deformable material hierarchy by introducing a PhysicsVolumeDeformableMaterialAPI. All deformable materials become independent and only inherit from PhysicsMaterial. Young's and Poisson's became unavailable to PhysicsSurfaceDeformableMaterialAPI and PhysicsCurvesDeformableMaterialAPI.

2. We found it less practical for implementations to map Young's and Poisson's parameters to deforming-mode stiffnesses.
 
LightWheels main argument is, if I understand correctly, that isotropic material properties are often easily available, while structural stiffnesses might be harder to find. 

Minimally we could provide the mapping of Young's and Poisson's to structural stiffnesses in the documentation (see attached PDF), or

1. Reintroduce common PhysicsDeformableMaterialAPI with Young's and Poisson's, and either keep PhysicsVolumeDeformableMaterialAPI as an alias or remove again.

2. Inline Young's and Poisson's in the surface and curves deformable materials.

sschirm:
Rest Shape Fallbacks, and Masses and Thicknesses on the Simulation Geometry
---------------------------------------------------------------------------------

Two updates have been pushed to this PR, one commit each.

**Rest shape fallbacks** (AOUSD Physics WG, 2026-07-08)

No rest shape attribute is mandatory. Where one is not authored, it is derived from the
corresponding attribute of the simulation geometry, taken at the default time code. The rules
are collected in a new Rest Shape Fallbacks subsection rather than spread through the per-type
paragraphs.

**Masses and thicknesses on the simulation geometry** (Deformables Discussion, 2026-08-12)

`surfaceThickness` and `curvesThickness` are removed from the two material APIs, and
`UsdPhysicsSurfaceDeformableSimAPI` and `UsdPhysicsCurvesDeformableSimAPI` gain a `thicknesses`
array. A thickness on the material confined that material to objects of a single thickness,
which defeats the reuse that binding one material to many objects is meant to provide.

`masses` and `thicknesses` are each paired with an `elementType` attribute in the array's own
namespace:

```
    # each elementType is constant, face, or point
    float[] physics:masses ()
    uniform token physics:masses:elementType = "" ()
    float[] physics:thicknesses ()
    uniform token physics:thicknesses:elementType = "" ()
```

The element type names the geometry features the values apply to, and the permitted set varies
by geometry: `constant`, `tetrahedron` and `point` for volumes, `constant`, `face` and `point`
for surfaces, `constant`, `curve`, `segment` and `point` for curves. Its fallback is the empty
string, a sentinel valid only for an empty array, so an array carrying values states the element
type of those values. Both arrays choose independently.

Primvars were dismissed in favour of the arrays and their `elementType` attributes, mainly
because no primvar interpolation addresses the segments of a linear `UsdGeomBasisCurves`, where
`uniform` is per curve, and `varying` and `vertex` are both per point. The namespaced form was
suggested by Matt Kuruc and seconded by Andrew Kaufman.

The volumetric fallback now reads its thickness from the geometry, and each stiffness table
states where its modes sample it: the face and the edge between two faces for surfaces, the
segment and the vertex between two segments for curves.

Mass Distribution is restructured around element masses. Every level of the precedence list
yields the mass of an element, and the lumping to points happens once, at the end, where a
simulator integrating at the points needs it. Nothing changes numerically. The per-point volume
decomposition over bound materials disappears with the rewrite, since each element simply takes
the density of the material bound to it.

**Two breaking changes for parser implementers**

1. `surfaceThickness` and `curvesThickness` no longer exist. Thickness is read from the
   simulation geometry, and may vary per element.
2. An authored `masses` array now requires its `elementType`. Per point was previously the only
   reading; the empty-string fallback means "not stated" rather than "per point".

proposals/physics_deformables/wp_deformable_physics.md
```diff
@@ -274,45 +266,82 @@ For volume deformables, a tetrahedral mesh is used to encode the simulation stat
...
     int4[] physics:restTetVertexIndices ()
-    float[] physics:masses ()
 }
 
 class "PhysicsSurfaceDeformableSimAPI"
 (
     inherits = </APISchemaBase>
 )
 {
+    # each elementType is constant, face, or point
```
andrewkaufman, line 292: @sschirm remind me, did we limit `PhysicsSurfaceDeformableSimAPI` to `UsdGeomMesh` explicitly? Otherwise if we allow other geom types to supply the "sim surface" the "face" elementType might not fit those...

proposals/physics_deformables/wp_deformable_physics.md
```diff
@@ -274,45 +266,82 @@ For volume deformables, a tetrahedral mesh is used to encode the simulation stat
...
+
+- `'constant'`: the whole geometry, in an array of size one. A single thickness value represents a constant value over the whole geometry. A single mass value represents the total mass, the same quantity that can be expressed by `UsdPhysicsDeformableBodyAPI:mass`.
+- `'curve'`: the curves of a `UsdGeomBasisCurves`. Each thickness value represents a constant value over its curve. Each mass value represents the total mass of its curve.
+- `'tetrahedron'`, `'face'`, `'segment'`: the elements of a `UsdGeomTetMesh`, `UsdGeomMesh` or `UsdGeomBasisCurves`.
+- `'point'`: the points of the simulation geometry's `points` attribute.
+
+Each `elementType` attribute has an empty string for its fallback value, a sentinel outside the supported element types. The empty string is valid only for an empty array, including an unauthored one.
+
+If a simulator requires an element type other than the authored one, the thickness at the desired location is obtained by averaging the adjacent samples, and masses are converted between points and elements as described in [Mass Distribution](#mass-distribution).
```
andrewkaufman, line 337: For thickness, did we consider linear interpolation vs averaging?

lyd405121:
@andrewkaufman Hi,andrew.
- Can I propose different material for curve GeomSubset
- Our artisit found it is diffcult to make a real cable matieral using single material



-  Every segment has its own material

image

SegmentSubgeomChain


## More
- And the physical attachment now is a ball joint , only limit position, but not rotation
- If we want to simulate a cable being weld to a rigidbody,  which means the linked segment shoud be fixed, it is now not supported by our schema
- Futher more, We have tested, three attachment modes:  ball, fixed, cable, the most stable one is cable joint,so it is better to support three kinds of attachment


https://github.com/user-attachments/assets/25a329e7-5602-4b5f-b6c7-5e73d0c46098

PR #113, opened 2026-09-11
Proposal for version branches in OpenUSD repository

sunyab:
### Description of Proposal

This document proposes introducing release branches in the OpenUSD repository for each release, instead of relying on a single `release` branch and tags on that branch to label releases. This addresses a problem where updates to the `dev` branch are blocked during the period leading up to an OpenUSD release due to the single `release` branch. It would also make managing patch releases easier, opening the door to more frequent security patches to older releases in the future.

[Link to Rendered Proposal](https://github.com/sunyab/OpenUSD-proposals/blob/release_branches/proposals/release_branches/README.md)

### Contributing

- [ X ] I agree to and accept the [Supplemental Terms](https://graphics.pixar.com/usd/release/contributing_supplemental.html).


sunyab:
Hey @nvmkuruc,

I think most of your questions below are about how we surface the OpenUSD version number in code. That's the subject of a separate proposal that we're hoping to publish soon. This proposal doesn't aim to change any of that, it's really just about the branch/tag structure in the repo.

> I wanted to ask about the "double zero" prefixing scheme for patches. It's been my experience that zero prefixing adds friction when converting to and from version schemes that require strict integers (ie., forgetting to pad when going back from integer to string). There's a list of "CalVer" versioning conventions [here](https://calver.org/users.html). Even tools that use zero prefixed months don't zero prefix their micro/patch release.

I originally read that GitHub's UI would sort tags lexicographically, so I added the zero-padding to ensure tags with patch releases would get listed in the right order. After some further digging it seems like this might not be the case -- for example, OpenImageIO's list of tags in the branch/tag dropdown are sorted numerically. I'll do some experiments, but I think we may be able to drop the leading zeroes in the tag name.

That said, at some point we'll need to include the patch version in the combined version value given by `PXR_VERSION` (or more likely a new macro, like `PXR_FULL_VERSION`). In that case, I believe we will need to reserve space for the patch version, e.g. `2605001` for v26.05.1. This'll be discussed in the other proposal I mentioned.

> Will the new version be included in the internal namespace?  We'd hope the answer would be "no" and that patches should maintain ABI compatibility.

I think the answer will most likely be as you say: patch versions should maintain ABI compatibility. But we haven't fully vetted this internally yet, so for now I'll just say TBD.

> Would you consider adopting "project version"? This helps cmake communicate project versions in a standard way and sets common project variables.  (Newer versions cmake can also communicate version compatibility.) https://cmake.org/cmake/help/latest/command/project.html#options
> 
> How will this surface in the `PXR{_MAJOR,_MINOR_,_PATCH,}_VERSION` strings?  If you adopted the `cmake` standards, could these be retired?

I haven't looked super deeply but off the top of my head I don't see any obstacles to specifying the OpenUSD version number in the cmake `project` call. The question is what we would specify as the major/minor/etc. versions for that call -- e.g., do we follow CalVer as we do in all of our external communications about OpenUSD version and use major=26, minor=8? Or do we use the SemVer values already published via `PXR_{MAJOR, MINOR, PATCH}_VERSION` and use major=0, minor=26, patch=8? That will be part of the proposal I mentioned above.

However, I think we do need to keep `PXR_{MAJOR, MINOR, PATCH}_VERSION` (or some form of those macros) around to make the version numbers accessible in code. AFAICT the `project` call makes those version numbers available to other CMake-based projects but not to C++ code itself. (

> Do you anticipate adopting the `SOVERSION` cmake property for libraries?

I believe setting an SOVERSION implies ABI compatibility guarantees, IIUC. If so, then I don't think we would set this now but probably would consider it if OpenUSD ever did provide those guarantees in the future.

PR #114, opened 2026-09-15
UsdSpatialMetricsAPI

tallytalwar:
### Description of Proposal

Here we discuss the introduction of `UsdSpatialMetricsAPI` as a successor to the `metersPerUnit` and `upAxis` layer metadata, along with a new `UsdGeomSpatialMetricsCompensationAPI` applied schemas.

[Link to Rendered Proposal](https://github.com/tallytalwar/OpenUSD-proposals/tree/varun/usdSpatialMetricsAPI/proposals/spatial-metrics-api)


proposals/spatial-metrics-api/README.md
```diff
@@ -0,0 +1,681 @@
...
+def Xform "World" (
+    prepend apiSchemas = ["SpatialMetricsAPI"]
+) {
+    double spatial:metersPerUnit = 0.1
+
+    def PhysicsScene "SceneRef" (
+        references = @physicsScene.usd@</Scene>
+        prepend apiSchemas = ["SpatialMetricsCompensationAPI"]
+    ) {
+        uniform token[] xformOpOrder = edit [
```
MishDivy, line 435: nit: `UsdPhysicsScene` inherits from `UsdTyped` not `UsdGeomXformable` perhaps not the best example. That said, the intent with prepending `metricsCompensation` xformaOp is clear and makes sense.

proposals/spatial-metrics-api/README.md
```diff
@@ -0,0 +1,681 @@
...
+not create or manage a separate compensation layer. Clients can however provide
+a compensation specific layer, like "MetricsAssembler Layer" NVIDIA's Metrics
+Assembler provides, which implements an automatic compensation example via a
+drag-and-drop listener or other asset importing hooks, and a background change
+listener, to update the compensation. Note that this can now use the utility
+methods provided by `UsdGeomSpatialMetricsCompensationAPI` to add these new
+`xformOps`.
+
+Compensation ops are added to `xformOpOrder` as a sparse array edit
+(`VtArrayEdit`, prepended via `VtArrayEditBuilder`), rather than restating the
```
MishDivy, line 306: **Question:** How do we handle the cases when there's a `!resetXformStack!` in the `xformOpOrder`? 

```
[xformOp:scale:metricsCompensation, !resetXformStack!, xformOp:someOtherTransform]
```

In the above, USD would ignore the `xformOp:scale:metricsCompensation` 

Perhaps the compensation utilities need to take  `!resetXformStack!`  into account?

proposals/spatial-metrics-api/README.md
```diff
@@ -0,0 +1,681 @@
...
+#### UsdSpatialMetricsAPI on subroot references
+
+`UsdSpatialMetricsAPI` may be applied at any prim, not only root prims.
+Sub-root references -- e.g. referencing a subcomponent model into a larger
+assembly -- are explicitly supported and not restricted. The inheritance behavior
+above, together with the spatial-metrics consistency validator (see 
+[Validators](#validators)),
+reconciles metrics at any depth in the hierarchy, so there is no requirement
+that the API live only on root prims.
+
```
MishDivy, line 199: **Question:**  In the case of a sub-root reference, how would the compensation utilities figure out the correct SpatialMetrics in the absence of `SpatialMetricsAPI` on the sub-root prims. For instance:

Pseudo Scene_1.usda 
```
/Machinery (Default Prim) (Has SpatialMetricsAPI: meterPerUnit = 0.01)
    Conveyor 
```

In the above, validators would pass as `/Machinery/Conveyor` would simply inherit the SpatialMertics from its parent's prim.

Pseudo Scene_2.usda
```
/World (Default Prim) (Has SpatialMetricsAPI: metersPerUnit = 1)
    /ConveyorRef (references @Scene_1.usd@</Machinery/Conveyor>)
```

In Scene_2.usda, if the client tries to call the `ApplyAndCompensate` then:
-  would it fail as the `/World/ConveyorRef` prim doesn't have `SpatialMetricsAPI`?  
   - This is not ideal for clients. And authoring duplicate API on `/Machinery/Conveyor` would cause redundancy and sync issues.
- or would it try to figure out the original authored `SpatialMetricsAPI` data applied on the `/Machinery` prim in `Scene_1.usda`
  - This may be more convenient for clients, but opening a layer, then tracking the ascendant prim with valid `SpatialMetricsAPI` to read the properties, does not seem to be the best approach either.

proposals/spatial-metrics-api/README.md
```diff
@@ -0,0 +1,681 @@
...
+The `xformOp:scale:metricsCompensation` reconciles the geometry -- the
+`/World/SceneRef/Ground` `Mesh` prim's `points` and any transforms compose
+correctly in the referencing stage's units. But `physics:gravityMagnitude` is not
+transformable, no `xformOp` can fix it. Because
+`UsdGeomSpatialMetricsCompensationAPI` is applied to the prim, physics clients /
+`usdPhysicsParser` can key off that marker to attach their own compensation
+behavior -- scaling `gravityMagnitude` appropriately for the differing
+`metersPerUnit`. A single schema application addresses both the transform domain
+(via the builtin `xformOps`) and other spatially-aware domains (via the marker).
+
```
MishDivy, line 452: **Question/Suggestion:** Would it make sense to allow schema authors to register custom compensation handlers that these utilities invoke?

For example, a studio might have a custom schema:

```
class Conveyor "Conveyor"
{
    # Acceleration in the asset's distance units per second squared.
    float maxAcceleration
}
```
Like the `gravityMagnitude` example, `maxAcceleration` would need conversion when the source and target distance units differ.

Could the schema author implement a "ConveyorCompensationHandler" and register it through OpenUSD’s plugin system? The compensation utilities in `UsdGeomSpatialMetricsCompensationAPI` could then discover and invoke the applicable handlers.

This would let applications that invoke these utilities use the same schema-specific compensation logic, provided the corresponding plugin is available. It could also provide a common mechanism for properties such as UsdPhysics’s `gravityMagnitude`.

I noticed there's precedence for a registration-and-dispatch pattern in `UsdGeomRegisterComputeExtentFunction`.

MishDivy (commented):
Thanks for this proposal! I am excited to see OpenUSD address this issue! ٩(^ᗜ^ )و ´-

Left some comments and questions.

proposals/spatial-metrics-api/README.md
tallytalwar: Good catch. You are absolutely correct that `PhysicsScene` does not inherit from `UsdGeomXformable`, so `xformOps` authored on it are not evaluated by `UsdGeomXformCache` and would have no effect on child geometry. My aim was the show the spatially aware attributes but non transforming and I overlooked this, so thanks for catching this.

I'll update the example to use an `Xform` parent for the reference so the compensation `xformOps` are on an `Xformable` prim and compose correctly to the children.

proposals/spatial-metrics-api/README.md
tallytalwar: Good question. 

We think its even more subtle than have the `!resetXformStack!` on the prim where compensation is applied, as `!resetXformStack!` could be authored on a descendant prim where compensation is authored on an ancestor, which will be much more convoluted to handle. Given this, we believe `!resetXformStack!` is an intentional authoring decision for the asset. If a prim resets its transform stack, it is deliberately opting out of inherited parent transforms, including any compensation that may exist on an ancestor.

I initially thought we could insert a compensation after the reset, via sparse edits, but **a.** inserts at a specific array index via sparse edits are fragile (the index can shift if other edits compose) and **b.** one will not be able to handle resets on a descendant prim via this mechanism. The compensation utilities will use `prepend` via a sparse `VtArrayEdit` and will not attempt to work around `!resetXformStack!`.

proposals/spatial-metrics-api/README.md
tallytalwar: In this scenario, `/World/ConveyorRef` references `</Machinery/Conveyor>` which doesn't have `SpatialMetricsAPI` applied, only its parent `/Machinery` does, and that parent is not part of the sub-root reference.

`ComputeEffectiveMetersPerUnit` on `/World/ConveyorRef` walks the composed hierarchy upward, it would find `/World`'s `SpatialMetricsAPI` (`metersPerUnit = 1`), which is the target, not the source. The source metric (what the asset was originally authored in) is not recoverable because the sub-root prim doesn't carry it and the parent's API from `Scene_1.usda` is lost across the reference arc.

If a sub-root prim is intended to be independently referenceable, `SpatialMetricsAPI` should be authored on it by the asset author. The proposal's sub-root references section already states: 

> "It is recommended for DCC / clients to author UsdSpatialMetricsAPI on prims likely to be referenced."

That said, this is a real gap in validation. We can add a validator that checks prims with authored references (`prim.HasAuthoredReferences()`) and warns if the prim does not have `SpatialMetricsAPI` directly applied (only inherited from the referencing context, not carried by the asset). The validator could also provide a fixer: it can inspect the referenced layer to recover the source prim's effective `SpatialMetricsAPI` values, then apply the SpatialMetricsAPI on the referenced prim with those values and apply the appropriate compensation accordingly.

proposals/spatial-metrics-api/README.md
tallytalwar: Thanks

I think, we intentionally keep the compensation API as a marker and not a dispatch mechanism. Domain-specific consumers that recognize the marker should apply their own attribute-specific compensation at consumption time rather than rewriting authored values in USD. For example, a physics parser that reads USD physics scene initial state descriptions into a simulator's format would be the natural place to apply physics-domain compensation. It reads the authored `gravityMagnitude`, sees that `UsdGeomSpatialMetricsCompensationAPI` is applied, computes the scaling factor from the effective metrics, and feeds the scaled value to the simulator. The authored USD value stays untouched; the compensation happens at the domain boundary. 

USD can provide a utility that computes the compensation factor from the source and target metrics, but the actual application to domain-specific attributes is the consumer's responsibility, each domain knows its own attributes' dimensional semantics better than a generic handler registry could express.

For the analogy with `UsdGeomRegisterComputeExtentFunction`, that registration provides the interface for plugins to register, but the computation itself is done by the plugin author for their specific type. Similarly here, each domain would implement its own compensation. We don't think a formal registration mechanism is needed and the marker plus the effective-metrics query utilities are sufficient for any domain client to implement compensation.

proposals/spatial-metrics-api/README.md
tallytalwar: Updated with [e8191ac](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/114/commits/e8191ac9be242e23e2e1075724e4050df735060b) to cater to just intended physics's spatially aware properties behavior attachment via the compensation API.

proposals/spatial-metrics-api/README.md
tallytalwar: Made a note about `resetXformStack` and its intended behavior and compensation not doing anything extra with it. [e8191ac](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/114/commits/e8191ac9be242e23e2e1075724e4050df735060b)

proposals/spatial-metrics-api/README.md
tallytalwar: Added a new validator for this: [e8191ac](https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/114/commits/e8191ac9be242e23e2e1075724e4050df735060b)

nvmkuruc:
Hi @tallytalwar.  I wanted to ask about the proposal's choice of using `prepend` to author correctives.  That suggests to me that you're expecting transforms for an asset composed into assembly will always be authored in the units of the asset, not the assembly.

If the resolved value of `xformOpOrder` is--

```
["xformOp:rotateX:metricsCompensation", "xformOp:scale:metricsCompensation", "xformOp:translate", "xformOp:rotateXYZ", "xformOp:scale"]
```

--the compensation ops affect the other ops.  If the assembly is declared to be meters, and the asset was declared to be centimeters, the translation would be authored in centimeters.  Is that the intent?

Are you worried that this might be confusing when rotational compensation is applied?  Tooling and users would have to be aware that a translation in Y would really be a translation in Z.

tallytalwar:
@nvmkuruc very valid point and thank you for bringing this up. @spiffmon and I had more discussion on this and feel maybe `VtArrayEdit` might not be a feasible solution here. Even with an `append` array edit, any subsequent `AddXformOp` call on the assembly side will result in these Ops applied more locally before any compensation xformOps are applied.

I will be posting another commit here soonish once we have an alternate figured / finalized for this issue. Thanks again for raising this.

nvmkuruc:
@tallytalwar Yeah. It's unfortunate that neither side is universally the right replace to inject compensation.

